PR #53, opened 2014-11-14
Adds ability for users to have multiple roles.

ben-github:
Adds new mapping table to link users with multiple roles. Tested and works, however, because it changes the ab_user table (removes the role_id column) and needs another table for role mapping, it breaks FAB when used with already existing database.

This is not ready to be merged in yet, but I did a pull request to keep comments and code in a central place. Hopefully I can add migration stuff and then we can do a merge if everything looks good. 

Migration should be easy (drop column, add mapping table and do insert based on user's role_id) but I'm not sure where / how to put that.


ben-github:
Just wanted to check in on this. Once this two issues get resolved, I'll rebase against master and add documentation and migration scripts so this can get added to a release.

dpgaspar:
Sorry i'll probably change this, on next version i'm adding the ability to use your own SecurityManager.
You can do it any way you like, there is one AbstractSecurityManager with the must have methods (abstract) and BaseSecurityManager with partial security logic backend agnostic.

Next version will partially support MongoDB also.

I think this is probably the best approach. If you want to, we can discuss including your SecurityManager on the release has an option, without the need to migrate DB. next release is branch 1.2.2 but will be 1.3.0. It will have some breaking features related to Filters.

dpgaspar:
No i it's not possible to enable multiroles without db change, i had a diferent idea, but it's not possible to do without some serious python hacking (and breaking python's zen).

Your right! the best way is to enable multiroles, my reluctance is about introducing too many breaking changes on next release.

i have a working multirole security for SQLAlchemy and MongoDB (but have not pushed it yet), i am thinking about introducing some new security features also:

- Copy role action
- Role generation for certain view, if choosen on UI will automaticaly create a role with all the required features (can list, can show, menu access etc) for a certain view.

Do you have some ideas on this also?

If you agree, i'm going to:
- Close this pull request.
- Push new working multirole version to new branch 1.3.0.
- You submit a new pull with a migration script (bullet proof) and docs to migration.

Thank you, once more.

ben-github:
Couple thoughts on role features:

* Copy role action is nice, but not as useful with multiple roles as you can assign people to the role you want to copy. That said, since there are so many permissions, it is not a bad idea.
* There are many roles with each view. Often that granularity is not needed. I'm thinking there could be a few meta-roles for each view: admin which has full access. And read_only which gets menu, and view, but no add, delete etc. Maybe a third role for usage: read_only plus add, but no edit / delete.  Then you can quickly dish out proper permissions without having to add multiple roles per view.

Sounds like a plan. Can you hold off closing this merge request until you push the  new 1.3.0 branch -- then I'll know when it is ready and put comments, pull requests there. 

Thanks!

dpgaspar:
I've just pushed 1.3.0. 

Thank you, once more.

ben-github:
Ok, I'll start poking around. Closing this pull request.

PR #85, opened 2015-01-22
First pass at migration script

ben-github:
I need (well, want) to use FAB's ability to add a table rather than writing raw SQL code since that code depends on the db type. I was able to run db.create_all() but in order to get that it work right, I had to manually type in  the SQLALCHEMY_DATABASE_URI since the config read had issues with paths. Maybe you can take a look as it would be nice to just run
```
python /path/to/bin/migrate-db_1.3.py
```

Tested with sqlite. Needs testing on mysql and postgresql.

Also I left a lot of the structure  / functions from the older 0.7 migration script. Maybe they will be useful in later migration scripts. I could remove them. Thoughts?

Also, I'm not deleting the role_id column on the user table as I think you mentioned wanting to keep it even though it isn't being used for compatibility reasons. If you want it removed, let me know and I'll add that as well.

### DIFF
--- a/bin/migrate_db_1.3.py
+++ b/bin/migrate_db_1.3.py
@@ -69,9 +69,29 @@ conn = engine.connect()
 log.info("Database identified has {0}".format(conn.engine.name))
 check_engine_support(conn)
 
+db.session.remove()
 db.create_all()
 
 conn.execute("INSERT INTO ab_user_role (user_id,role_id) SELECT id,role_id from ab_user")
+
+if conn.engine.name == 'postgresql':
+    sequenceremap={ 'seq_ab_permission_pk':'ab_permission_id_seq',
+                    'seq_ab_view_menu_pk' :'ab_view_menu_id_seq',
+                    'seq_permission_view_pk': 'ab_permission_view_id_seq',
+                    'seq_ab_permission_view_role_pk': 'ab_permission_view_role_id_seq',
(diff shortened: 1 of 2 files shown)


dpgaspar:
Some remarks:

Dont' use::

     print "bla, bla" # it's not python 3, compat

use::

     print("bla, bla")

even better::

     log.info("bla, bla")

Please do not use INSERT if you can use SQLAlchemy. INSERT bla SELECT bal will fail on some engines.
Fetch all users, loop, append role into roles.

Could you please submit the ALTER SEQUENCE NAME for postgre and ORACLE?

ben-github:
Switched from print to log.info. For using SQLAlchemy instead of direct SQL, how do I get access to the SQL tables in SQLAlchemy in this script. I have app and db but I'm not seeing how to reference the SQLA instance of the User / Roles tables, etc. 

I'll add that sequence rename for PG / Oracle. I think that will have to be raw SQL.

dpgaspar:
Exactly raw SQL, there is no other way, without using an external migration package.

The User/Role mig prob. don't worry about that now. I'm going to include this script inside fabmanager command line (version 1.3.0, flask_appbuilder/console.py). Submit the alter sequence.

Thank you, once again for your help.

ben-github:
Just pushed a change that handles renaming sequences in postgres and gives the raw SQL in the documentation.

dpgaspar:
Thanks!

PR #115, opened 2015-03-10
bugfix. fix 'multi line confirmation' not display

muxueqz:

PR #124, opened 2015-03-24
Revert "bugfix. fix 'multi line confirmation' not display"

dpgaspar:
Reverts dpgaspar/Flask-AppBuilder#115


dpgaspar:
New version 1.3.5 is out. You can use HTML on the action text.

Issue #138, opened 2015-04-04
shell option for fabmanager


lu-zero:
While trying something I noticed that fabmanager does not provide a shell option, would be useful just for me?

lu-zero:
I implemented it [here](https://github.com/dpgaspar/Flask-AppBuilder/pull/139)

dpgaspar:
Pull request will be on hold, until Flask 1.0 is out.

PR #139, opened 2015-04-04
Shell command

lu-zero:
It mimics Flask-Script `shell` command.

### DIFF
--- a/flask_appbuilder/console.py
+++ b/flask_appbuilder/console.py
@@ -106,6 +106,39 @@ def run(app, appbuilder, host, port, debug):
     _appbuilder = import_application(app, appbuilder)
     _appbuilder.get_app.run(host=host, port=port, debug=debug)
 
+@cli_app.command("shell")
+@click.option('--app', default='app', help='Your application init directory (package)')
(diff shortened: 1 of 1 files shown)


dpgaspar:
This looks great and it's a good idea also, but it will only work with Flask 1.0 (dev), this is not out yet.

I could merge it already, but it won't work for the majority, latest version is 0.10.1, and there is no **make_shell_context**. #138

lu-zero:
Oh, right, I forgot about it. Another item that might be nice is integrate alembic like it is done in Flask-Migrate.

dpgaspar:
That would be nice too, are you willing to submit a pull for it?

lu-zero:
I'll try this week-end it isn't as simple as for the shell and I have quite a bit of work to do.

dpgaspar:
@lu-zero, sorry closing this one. 

On the future fabmanager should be converted to the "new" flask shell.

Something like:

```
flask fab create-admin
....
```
or
```
flask fabmanager create-admin
....
```

lu-zero:
It is fine, sadly I got way too busy with other tasks =/

PR #155, opened 2015-04-17
New LDAP options - bind to server before search and allow self signed certs

a-mckinley:


dpgaspar:
This looks great, thanks, i'll run some tests and integrate it on the next version.

PR #160, opened 2015-04-23
replaced render_template with self.render_template

mortbauer:
I made two small change since it failed to render the activation template since base_template was undefined, works fine now, I also changed it for the rendering of the email template, although this didn't make any problems, but this way it is always the same rendering function.

The second change in the same file is from get_role_by_name to find_role, since get_role_by_name does not exist anymore 

greets martin


dpgaspar:
Thanks, there are some more things to update to get the registering of users to work.

Were you able to make it work with this change?

PR #178, opened 2015-06-13
my fix to the multiview bug with charts... this fixes #177

### LINKED ISSUE #177: MultipleView javascript bug with 2 (or more?) charts
if one tries to build a multiple view with two charts (GroupByChartView) two "slots" (divs) are rendered into the page but only one chart is plotted.

it seems that all charts are rendered into the <div id="chart" ></div>, therefore all charts are overlapping due to the fact that many divs now have the same "chart" id and only the first is used by javascript functions... 
also the id="chart" thing is hardwired all alround even into the javascript function responsible of the creation of the charts: function createChart().
I think the solution is to assign a progressive id to divs, like id="chart0", id="chart1" and so on and adjust the createChart() function, but I've not investigated more...

absinthetized:
as per title


dpgaspar:
Thanks i'll make some tests and include it on the next release 1.4.1. Coming out soon.

PR #179, opened 2015-06-14
minor fix to display proper chart title.

absinthetized:
previous bugfix was a bit aggressive in substituting spaces with underscores.
this commit restores a properly rendered chart title on screen, respecting
original spaces.


dpgaspar:
Thank you, i have already noticed that, i've changed 2 things from your previous pull request:

1 - Added modelview_name to the chart widgets
2 - Changed {{ chart_title|replace(' ','_') }} to modelview_name.
3 - title is chart_title

You can take a look at: https://github.com/dpgaspar/Flask-AppBuilder/blob/1.4.1/flask_appbuilder/templates/appbuilder/general/widgets/direct_chart.html

This was i nice catch, thank you again for reporting and solve it!

PR #193, opened 2015-07-21
added a number of new filters for the generic interface, plus a fix to the 'like' and 'equal' operators

absinthetized:
hi,

due to my needs I've derived a series of generic* classes into my project and I've implemented additional filters for the generic interface. I don't know if this can help the project in anyway.

The expected behaviour is to have out-of-the-box filters for equal, greater, smaller, like, insensitive like, starts with and so on...

In order to make pull request a bit more streamlined, I've merged the contents of my custom classes into the FAB's base classes I've derived from.

The merged code needs *a lot* of debugging as I've merged it very fast in order to avoid to put a lot of crap into my custom classes making future merges impossible.

hope this helps.

bye,
MN

### DIFF
--- a/flask_appbuilder/models/generic/__init__.py
+++ b/flask_appbuilder/models/generic/__init__.py
@@ -176,12 +176,85 @@ class GenericSession(object):
     #           FUNCTIONS for FILTERS
     #-----------------------------------------
 
+    def starts_with(self, col_name, value):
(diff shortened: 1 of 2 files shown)


dpgaspar:
This looks great @absinthetized, thanks. i'll merge and run some tests. It should be included on the next version 1.4.4, coming out soon

PR #204, opened 2015-08-07
Login button text fix

Fuzzwah:


dpgaspar:
Seems right, thank you. I'll run some tests and include it on the next release.

PR #206, opened 2015-08-07
registration form fields aren't being validated

Fuzzwah:
I was able to register users with the same usernames,


dpgaspar:
I'll merge and run some tests. Will be included on the next version.

PR #209, opened 2015-08-08
Made several changes to query method:

mgellesch:
- Greatly improved overall speed
- len(objs) previously retrieved all objects. The recommended approach is
  objs.count(). Greatly reduces overhead.
- Added some error checking and a simple warning if lots of data is being
  downloaded
- Fixed error: if page given but not page_size, None error when multiplying,
  line 47 (old version)
- Made it slightly more pythonic


dpgaspar:
Great work! thank you, i'll run some tests and include it on the next version

PR #211, opened 2015-08-09
UTF-8 encoding for the json label strings

### LINKED ISSUE #210: UnicodeEncodeError is rised on API routes when using unicode characters
In my labels I have latin characters "č,ć,š..." and this error is rised when I try to access for example `users/api/read` route

```
UnicodeEncodeError

UnicodeEncodeError: 'ascii' codec can't encode character u'\u010d' in position 7: ordinal not in range(128)
```

mirzap:
Fixes #210


dpgaspar:
I've noticed this bug, let me run some tests. Thank you, this will be included on the next version.

Issue #216, opened 2015-09-01
English issues found during translation


michalfita:
I've done Polish translation today for Flask AppBuilder but I found couple major issues with the English used in original strings:

* in `security/forms.py:10` there is "openid" all small caps like identifier not a name of the feature, so I think this should be "OpenID";

* in `security/forms.py:12` there is "remeber_me" that again looks like identifier, but rather should be descriptive text "Remember me";

* in `security/views.py:12` there is "Its" while should be "It's" - very common mistake;

* in `security/views.py:213` there is "users" while should be "user" or "user's" as plural form doesn't match there;

*  in `templates/appbuilder/general/security/register_oauth.html:15` there is "Signin" while I think should be "Sign in" (lack of whitespace).

These are only major language issues I spotted, there is lot more minor issues related to grammar and pronunciation of English, for example there is no clear distinction between general text and message texts, which should be sentences ended with a dot (or full stop as say in England). I could fix these according to my knowledge (I'm not a English native speaker) but that would require fixing all translations as well. But I'm sure with these original English improvements it would be easier to translate. The only thing is that I would make it more British English, no US English (which again would make it easier to translate).

dpgaspar:
Yes, your right, thanks! Are you willing to submit a PR for this?

dpgaspar:
Thank you once more!

PR #218, opened 2015-09-02
Issue #216 English issues found during translation

michalfita:
This commit contains fixes for original English strings found during
translation to Polish. This, however, is not a full blown attempt to
fix English in whole project, as this requires more work and more
careful approach to the context.


dpgaspar:
Looks like great work, Thanks for this!

PR #219, opened 2015-09-03
Making the inline crud stateless

mistercrunch:
This is work in progress, I need guidance. 

I feel like this almost but there's some magic behind the scene that I don't understand...

### DIFF
--- a/flask_appbuilder/views.py
+++ b/flask_appbuilder/views.py
@@ -565,18 +565,22 @@ class MultipleView(BaseView):
                                     views_widgets=views_widgets)
 
 
-
 class CompactCRUDMixin(BaseCRUDView):
     """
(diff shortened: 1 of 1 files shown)


mistercrunch:
@dpgaspar, this works now and the PR is ready for review. I tested it against a 8 thread server and I don't see the problems I used to see.

dpgaspar:
Looks great, i will merge it, test it some more, and include it on the next version.

Thanks!!

mistercrunch:
Awesome, it seems to have fixed all the issues I had on my multi-threaded server.

BTW if you are interested to see what I'm building on top of FAB here's a link to my repo:
https://github.com/mistercrunch/panoramix

The project is still young (started at a hackathon in July) but I think I may be spending more cycles on it starting soon.

PR #223, opened 2015-09-09
Redoing the undone - Proxy support

mistercrunch:
Somehow I overwrote my original fix with my most recent one. This puts proxy support back into place.


dpgaspar:
Thanks once more.

PR #241, opened 2015-10-14
fixed small bug in generic filters where both like and ilike shown same desc

absinthetized:


dpgaspar:
Thanks, will be included next version 1.4.8 (or 1.5.0). I'll have to make translations for this one...

PR #247, opened 2015-10-16
fixed a bug when ordering view columns where None values are in.

absinthetized:
patch coming from a stackoverflow snippet


absinthetized:
now it doesn't chrash anymore but sorting is gone actually, and randomness is in... let debug it a bit more...

absinthetized:
ok, this time it should be ok!

dpgaspar:
Thank you, once more, i'll run some tests, and if all ok, will be included on next release.

PR #284, opened 2016-01-11
Allowing apps to alter title using a jinja block

mistercrunch:
I'd like to alter the title in my app.


dpgaspar:
Thanks, i'll merge it and run some tests.

PR #285, opened 2016-01-13
updated naming of a few filters

Fuzzwah:
Maybe I'm being a bit pedantic, but I feel my changes are better names for the filter options


Fuzzwah:
I missed one, closing and I'll make a new pull request

PR #286, opened 2016-01-13
Filter names for date fields; before, after

Fuzzwah:
Rather than "Greater than" and "Less than", filters for date based fields should read "After" and "Before".


Fuzzwah:
I made a bit of a mess with these commits. Going to close this request and do it fresh.

PR #287, opened 2016-01-13
updated naming of a few filters

Fuzzwah:
Added "Yes" and "No" for bool fields, "Before" and "After" for date and datetime fields.

Updated a few others to read better (imho).


dpgaspar:
Great, this is an improvement. thought it will probably "break" i18n, and worse it may break many implementations. Have to run some tests and think about it.

dpgaspar:
duplicate of #327

PR #300, opened 2016-02-17
added datetime.date objects to generic models

absinthetized:
now the generic model/session is able to cope with datetime.date objects. While the _equal operator has only been tested filling generic models with dates coming from a sqlite db, it seems that SQLAlchemy makes a good effort in standardiziing the format of the date objects output, so generic models should be OK with other backends.

*ANYWAY* proper test with other db is suggested. feel free to drop this pull if you can't trust it: In the end I've not figured out how to test unit the thing!

### DIFF
--- a/flask_appbuilder/models/generic/__init__.py
+++ b/flask_appbuilder/models/generic/__init__.py
@@ -3,7 +3,7 @@ __author__ = 'dpgaspar'
 import operator
 import os
 from ..._compat import with_metaclass
-from datetime import date
+from datetime import date, datetime
 
 #--------------------------------------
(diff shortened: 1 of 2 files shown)


absinthetized:
forgotten to add code for the remaining filters :-(
I'm aging in a really bad way...

will be back asap with the full request, sorry for polluting your github page!

absinthetized:
ok added filters. all of them.
adding here that, while no unit test is present, I'm beta testing the stuff in my product right now. and my cat is still alive :P

dpgaspar:
Thanks, i'll merge it and test it.

Issue #314, opened 2016-04-14
Issues with flask app builder LDAP with AD


mogthesprog:
Hi I'm having some trouble configuring LDAP AUTH with AD for Caravel. I'm not sure whether this is an issue with appbuilder or caravel.

I get the following output when trying to authenticate:

```
2016-04-14 12:17:39,686:INFO:werkzeug:10.192.137.33 - - [14/Apr/2016 12:17:39] "GET /static/assets/images/favicon.png HTTP/1.1" 200 -
2016-04-14 12:17:45,932:DEBUG:flask_appbuilder.security.manager:LDAP indirect bind with: gblsci_ldapsrch@dunnhumby.co.uk
2016-04-14 12:17:45,936:DEBUG:flask_appbuilder.security.manager:LDAP BIND indirect OK
2016-04-14 12:17:45,937:DEBUG:flask_appbuilder.security.manager:LDAP got User [('CN=Morgan Jones,OU=Solutions Architecture,OU=Information Technology,OU=Data Solutions,OU=Standard,OU=Users,OU=UK,DC=dunnhumby,DC=co,DC=uk', {'mail': ['Morgan.Jones@dunnhumby.com'], 'givenName': ['Morgan'], 'sn': ['Jones']}), (None, ['ldap://ForestDnsZones.dunnhumby.co.uk/DC=ForestDnsZones,DC=dunnhumby,DC=co,DC=uk']), (None, ['ldap://DomainDnsZones.dunnhumby.co.uk/DC=DomainDnsZones,DC=dunnhumby,DC=co,DC=uk']), (None, ['ldap://dunnhumby.co.uk/CN=Configuration,DC=dunnhumby,DC=co,DC=uk'])]
2016-04-14 12:17:45,937:DEBUG:flask_appbuilder.security.manager:LDAP bind with: CN=Morgan Jones,OU=Solutions Architecture,OU=Information Technology,OU=Data Solutions,OU=Standard,OU=Users,OU=UK,DC=dunnhumby,DC=co,DC=uk ********
2016-04-14 12:17:45,942:DEBUG:flask_appbuilder.security.manager:LDAP bind OK: CN=Morgan Jones,OU=Solutions Architecture,OU=Information Technology,OU=Data Solutions,OU=Standard,OU=Users,OU=UK,DC=dunnhumby,DC=co,DC=uk
```

Below is my config:

```
AUTH_TYPE = AUTH_LDAP

[...]

PR #319, opened 2016-05-05
Fixes #273 by not registering a view that will not exist for LDAP

### LINKED ISSUE #273: Help needed adding new users automatically with LDAP authentication.
By following the process outlined in #158 ,(i.e. creating an admin user with credentials that are the same as one which can be looked up via LDAP), I have managed to get LDAP authentication working for an admin Role, though in the docs it suggests that other users will be automatically added to the user table in the database and this does not seem to happen "out of the box". I was wondering if there's something that I've missed. Here is what I have tried so far:

## Using Setting in the Config File 

In the config I tried setting the following:

```
AUTH_ROLE_PUBLIC = 'Public'
AUTH_USER_REGISTRATION = True
```

Though then encountered the error:

```
>>> python run.py
2015-11-09 15:46:29,759:INFO:flask_appbuilder.security.sqla.manager:Security DB not found Creating all Models from Base
2015-11-09 15:46:30,914:INFO:flask_appbuilder.security.sqla.manager:Security DB Created
2015-11-09 15:46:31,017:INFO:flask_appbuilder.security.sqla.manager:Inserted Role: Admin
2015-11-09 15:46:31,141:INFO:flask_appbuilder.security.sqla.manager:Inserted Role: Public
2015-11-09 15:46:31,144:WARNING:flask_appbuilder.security.manager:No user yet created, use fabmanager command to do it.
2015-11-09 15:46:31,147:INFO:flask_appbuilder.base:Registering class IndexView on menu
2015-11-09 15:46:31,272:INFO:flask_appbuilder.base:Registering class UtilView on menu
2015-11-09 15:46:31,392:INFO:flask_appbuilder.base:Registering class LocaleView on menu
2015-11-09 15:46:31,529:INFO:flask_appbuilder.base:Registering class NoneType on menu
Was unable to import app Error: 'NoneType' object has no attribute 'appbuilder'
```

On tracing this back it seems there is no register user view for LDPA defined by default, and this causes the following line to return 'None' 
https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/manager.py#L391

Do I need to define my own register user view for LDAP?

bibby:
Without `AUTH_USER_REGISTRATION` set to `False`, ldap users logging in for the first time are bound successfully, but an account is never made for them (See #314).

Enabling `AUTH_USER_REGISTRATION`, however sees the `SecurityManager` attempting to load a view that will never come. This change loads the view only when there is one to be had.

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -388,7 +388,8 @@ class BaseSecurityManager(AbstractSecurityManager):
                 self.registeruser_view = self.registeruseroidview()
             elif self.auth_type == AUTH_OAUTH:
                 self.registeruser_view = self.registeruseroauthview()
-            self.appbuilder.add_view_no_menu(self.registeruser_view)
+            if self.registeruser_view:
+                self.appbuilder.add_view_no_menu(self.registeruser_view)
 
         self.appbuilder.add_view_no_menu(self.resetpasswordview())
         self.appbuilder.add_view_no_menu(self.resetmypasswordview())


dpgaspar:
Thanks, i'll merge it and test it. I'm planning in doing some rework on this. My plan is to add a registration form for LDAP, i think this will be a more expected behaviour, all auth type on self registration have a registration form, so everything works the same way.

PR #327, opened 2016-05-24
filter names

Fuzzwah:


dpgaspar:
@Fuzzwah i agree, the english filter name are not great.

I'll merge this later, this would impact all the translations.

Could you further explain the FilterYes and FilterNo ?

PR #331, opened 2016-06-01
fix user not being created when logging in through an OAuth provider

deniszgonjanin:
When using an OAuth provider, no user is created after a successful provider authentication. The user is ejected right away and the login fails. This PR changes that so that a new user is created, just like with other providers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ app.db
 env
 venv
 *.sublime*
+*.egg-info/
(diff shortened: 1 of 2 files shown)


dpgaspar:
The only auth method that creates the user right away is LDAP, all the other direct the user to a registration form. Also you need to set AUTH_USER_REGISTRATION = True.

If you don't use that, you need to first create the user on the DB, then they can login.

PR #332, opened 2016-06-06
Realign deprecated references to `flask.ext`

mistercrunch:
From: http://flask.pocoo.org/docs/0.11/extensiondev/#extension-import-transition
```
As of Flask 0.11, most Flask extensions have transitioned to the new
naming schema. The flask.ext.foo compatibility alias is still in Flask
0.11 but is now deprecated – you should use flask_foo.
```


dpgaspar:
Great, thanks!

PR #340, opened 2016-06-17
Update Chinese Translations

rocky1001:
Update translations as a Chinese native speaker :)


dpgaspar:
Thanks @rocky1001,

I'll trust this will make it better. :)

PR #348, opened 2016-06-26
Flask extension proper import

anddam:
This replaced all occurencies of flask.ext.something with flask_something to avoid the deprecation warning in Flask 0.11

http://flask.pocoo.org/docs/0.11/extensiondev/#extension-import-transition


dpgaspar:
Ok, Thanks

anddam:
Oh btw, I replaced only occurencies in code, you may have to regenerate docs.

PR #358, opened 2016-07-10
Improvement to example of custom index view

dlocke:
Update the name of the template filename to match the name suggested earlier in the documentation.


dpgaspar:
right! Thanks

PR #360, opened 2016-07-21
Fix OAuth authentication when the OAuth provider returns an empty username

LAlbertalli:
Useful for Google Auth that returns an empty username when the user hasn't signed up on Google Plus


dpgaspar:
Sorry for the delay, can you rebase it?

PR #374, opened 2016-09-16
added default autosizing to app image (if any)

absinthetized:
this will prevent images to flood out of navbar estate.

I've recently added an icon for the app and the image has been rendered in its full size, going out of the navbar. this should fix the thing.
While I do not expect that _height="100%" width="auto"_ will be the one-fit-for-all, at least the default is now more sane IMHO.
User can override the template and customize it anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
   - "2.7"
   - "2.6"
 install:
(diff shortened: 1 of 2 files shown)


absinthetized:
ok merged my pip-deprecated branch just to be sure I've not broken any test...

mistercrunch:
LGTM

PR #375, opened 2016-09-16
fixed bug with pip --use-mirrors option (deprecated and now removed)

absinthetized:
travis is now broken due to pip changes. this should fix... let's try :-)


absinthetized:
OK coverage, got it! :-P

mistercrunch:
Already addressed in master, closing

PR #393, opened 2016-10-26
More sensible default page size

MichaelPereira:
Hi,

I would like to propose this change to the default page sizes for lists, which currently has too many little increments and a low maximum value:
- 5
- 10
- 15
- 20
- 25
- 30
- 35

Which leads users to always select 35 and still leaves them with high page counts if a view has a few hundreds items.

With the proposed change, the list will have less items with bigger increments:
- 25
- 50
- 75
- 100


mistercrunch:
LGTM

doctording:
page_size settings should be flexible, 
eg. a list of page_size

Issue #396, opened 2016-11-03
implementing OAUTH with Caravel


neevany:
Im trying to implementing google OAUTH with Caravel
using
Flask-AppBuilder==1.8.1
Flask-OAuthlib==0.9.3
but i am facing this error after redirecting from google

```py
flask_appbuilder/security/views.py", line 528, in oauth_authorized
return redirect(self.appbuilder.sm.registeruseroauthview.get_default_url(**userinfo))
UnboundLocalError: local variable 'userinfo' referenced before assignment
```

any ideas how i can fix this.

xrmx:
@neevany does the patch above works for you?

swatisinghi:
I am also seeing an issue with google oauth. Is there any working example specifically with google oauth that I can try. I tried using your patch as well but it did not help.

xrmx:
@swatisinghi if you have a stackstrace post it, otherwise it's quite hard to help you. I think there's people using oauth with google for authenticating with superset.

jigneshvasoya:
Same issue: adding stack trace here:

*******************************************************************************************************************
UnboundLocalError
UnboundLocalError: local variable 'userinfo' referenced before assignment

Traceback (most recent call last)
File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1836, in __call__
return self.wsgi_app(environ, start_response)
File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1820, in wsgi_app
response = self.make_response(self.handle_exception(e))
File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1403, in handle_exception
Display the sourcecode for this frameOpen an interactive python shell in this framereraise(exc_type, exc_value, tb)
File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1817, in wsgi_app
response = self.full_dispatch_request()
File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1477, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1381, in handle_user_exception
reraise(exc_type, exc_value, tb)
File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1475, in full_dispatch_request
rv = self.dispatch_request()
File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1461, in dispatch_request
return self.view_functions[rule.endpoint](**req.view_args)
[...]

PR #397, opened 2016-11-09
securty: don't crash on oauth errors

xrmx:
Don't reference userinfo if it has not been set.

Refs #396


mistercrunch:
LGTM. Thanks for the PR.

PR #402, opened 2016-11-23
Adding column_formatters to ModelView

mistercrunch:
Dictionary of formatter callables used to format columns in list and show widgets. This allows user to apply whatever formatting they want to their data. Similar to:
http://flask-admin.readthedocs.io/en/latest/api/mod_model/#flask_admin.model.BaseModelView.column_formatters

For context, I'm currently attempting to port the [Apache Airflow](https://github.com/apache/incubator-airflow) UI from `Flask-Admin` to FAB and may need to match some of the functionality there. The incentive is getting auth, and REST for all of our model from FAB while having a easy-ish port from Flask/ Flask-Admin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "2.7"
   - "2.6"
 install:
+  - "pip install pip --upgrade"
   - "pip install -r requirements.txt"
   - "pip install coveralls"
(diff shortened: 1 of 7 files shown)


dpgaspar:
Great, thks!

PR #409, opened 2016-12-21
Fixes for Google OAuth

vavrusa:
We're using FAB as a Superset dependency and wanted to use OAuth with Google.
This fixes use of deprecated Google+ API call and broken self registration.


songyanho:
@mistercrunch  Any update on this PR? Seems like the example code demonstrated the login with Google Oauth is not working

mistercrunch:

PR #413, opened 2017-01-11
Example masterdetail: bugfix

HardBite:
Hi!
I was not able to run masterdetail example on python 3.4 due to error:
`Was unable to import app Error: No module named 'models' `
so I changed import statement in views.py.

I also was unable to seed database with testdata.py due to error:
`TypeError: sequence item 0: expected str instance, bytes found`

I made quick small fixes, and hope they will be helpful.

### DIFF
--- a/examples/masterdetail/app/views.py
+++ b/examples/masterdetail/app/views.py
@@ -5,7 +5,7 @@ from flask_appbuilder.charts.views import ChartView, TimeChartView
 from flask_babel import lazy_gettext as _
 
 from app import db, appbuilder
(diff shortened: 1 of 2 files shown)


mistercrunch:

gallamine:
The `Was unable to import app Error: No module named 'models'` error is still present in the `quickcharts` example.

mistercrunch:
Last I checked `master` is fixed:
https://travis-ci.org/dpgaspar/Flask-AppBuilder/branches

PR #426, opened 2017-02-06
Fixes #425

### LINKED ISSUE #425: User bind after indirect bind fails when using userPrincipalName with MSAD
When using MSAD and the 'userPrincipalName' as the uid_field, `auth_ldap_append_domain` should not be added to the username returned by _search_ldap.

If using uPN, the username returned by _search_ldap is in the form `CN=User Name,...,DC=example,DC=com`. After this indirect bind and search, the application tries to append the domain again, which then makes it try to do a direct bind using `CN=User Name,...,DC=example,DC=com@example.com`, which is not a valid uPN.

NeilHanlon:
Probably not the cleanest way, but as long as we don't try to append
@domain.local to the uPN, it should work. Should be noted that using
sAMAccountName instead of userPrincipalName works as well.


NeilHanlon:
I'm pretty sure I did not cause any CI errors... can we re-run?

@dpgaspar

mistercrunch:

NeilHanlon:
@mistercrunch Well that's an annoying rebase... I can kill this and recreate a pull request with a clean history if you want.

Just let me know.

mistercrunch:
Looks like something bad happened please fix the change history or reopen the PR

PR #429, opened 2017-02-15
Update views.rst

robert-digit:
Updated flask_babel import since the old style no longer worked (and flask.ext.babel generates deprecation warnings)


robert-digit:
Pretty sure a one-line doc change shouldn't break CI.

PR #431, opened 2017-02-17
Log stack trace on error and prevent view menu creation bug

mistercrunch:
Many unexpected errors are silenced in a way that is hard to track down. Adding some `logging.exception(e)` in base.py will help make errors be more verbose and provide a way to track them down.

I went down the rabbit hole when a perm on a new view was not getting created as it should. Adding a simple condition helped moving forward, though I'm guessing the root cause is at least one degree removed.

### DIFF
--- a/flask_appbuilder/security/sqla/manager.py
+++ b/flask_appbuilder/security/sqla/manager.py
@@ -357,7 +357,7 @@ class SecurityManager(BaseSecurityManager):
 
     def exist_permission_on_views(self, lst, item):
         for i in lst:
-            if i.permission.name == item:
+            if i.permission and i.permission.name == item:
                 return True
         return False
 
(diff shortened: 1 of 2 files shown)


flask_appbuilder/security/sqla/manager.py
```diff
@@ -357,7 +357,7 @@ def del_permission_view_menu(self, permission_name, view_menu_name):
 
     def exist_permission_on_views(self, lst, item):
         for i in lst:
-            if i.permission.name == item:
+            if i.permission and i.permission.name == item:
```
mistercrunch: This is the bugfix. In my specific case `i.permission` was == None and `.name` was failing. I'm unclear on why a None permission made it here. Perhaps some incomplete deletion/cleanup?

PR #453, opened 2017-04-25
Added support for users to login with their username or email address

yon2004:
It's fairly simple change makes it easy to login.


MichaelPereira:
I agree that it would reduce the number of people complaining about not being able to login

yon2004:
As the user model requires email to be unique. Where would be the best place to check for registration email being unique?

mistercrunch:

PR #457, opened 2017-05-06
fix: dont overwrite csrf_token on form fill

S-YOU:
When CSRF_ENABLED (WTF_CSRF_ENABLED), there is error with User object don't have csrf_token.

This fixes https://github.com/airbnb/superset/issues/2232 for me, 
because csrf_token is overwritten to None.

Also, there is some deprecation warnings that already fixed on master, can we get minor version bump soon?


mistercrunch:
LGTM, @dpgaspar can we get a minor release?

PR #467, opened 2017-05-26
Fix two instances of urls being generated wrongly when running under a prefix.

joharohl:
This PR fixes two parts of flask-appbuilder that doesn't work when running a site under a prefix.

The most important one is in the CompactCRUDMixin used for inline add/edit views for a ModelView. The session parameter that is used to set the form action only used `request.full_path`. This will not work when the wsgi `SCRIPT_NAME` is different than `''`. Adding in the `request.script_root` solves this problem.

The second one is that the DBAuthView redirects to `/` rather than `self.appbuilder.get_url_for_index`. This will also break on a flask app build that uses a different url for it's index page than `/`

### DIFF
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -372,7 +372,7 @@ class AuthDBView(AuthView):
     @expose('/login/', methods=['GET', 'POST'])
     def login(self):
         if g.user is not None and g.user.is_authenticated():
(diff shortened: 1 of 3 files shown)


dpgaspar:
Thks!

joharohl:
No worries. Thanks for a great piece of kit.

PR #472, opened 2017-06-04
Security: Prevent masquerade attacks through oauth providers

mattk42:
This PR changes the ID that is used to link a google login to an account to an immutable user ID that is provided by google. The previous implementation used the display name, which could easily be changed at any point and used to log in to a different account by an attacker.

This also prefixes all oauth created users with the name of the provider that created it. This prevents someone from creating a colliding account through a different provider to gain access to an application.

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -378,7 +378,7 @@ class BaseSecurityManager(AbstractSecurityManager):
         if provider == 'google':
             me = self.appbuilder.sm.oauth_remotes[provider].get('people/me')
             log.debug("User info from Google: {0}".format(me.data))
-            return {'username': "google_" + me.data['id'].get('value',''),
+            return {'username': "google_" + me.data['id'],
                 'email': me.data['emails'][0].get('value',''),
                 'first_name': me.data['name'].get('givenName',''),
                 'last_name': me.data['name'].get('familyName','')}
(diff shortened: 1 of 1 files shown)


dpgaspar:
Hi @mattk42,

I've just merged #409 onto master that changes google OAuth, could you please update and resubmit with necessary changes?

This:
```python
me = self.appbuilder.sm.oauth_remotes[provider].get('people/me')
````
is now this:
``` python
 me = self.appbuilder.sm.oauth_remotes[provider].get('userinfo')
```

dpgaspar:
Sorry one last question, I may be mistaken but this is a breaking change, since usernames will change, and current deployed apps will not longer be able to allow logins. Is this correct?

mattk42:
I will take a look integrating with that other fix tonight or tomorrow.

You are correct that this will break existing users unfortunately.

dpgaspar:
We can add an extra column on ab_user table named 'provider'. If ab_user.provider is null we do not enforce username + provider check, yet all new users (self registration) would get this column filled.

This way we would not break currently deployed apps using OAuth. Though no security fix until provider column if properly filled.

Thoughts?

dpgaspar:
One more thing, thank you for your time @mattk42 👍  really appreciate it!

mattk42:
That sounds like a good idea too. I am not super familiar with flask as a
whole, so may take me a bit to get that in. But I can definitely take a
stab at it.

mattk42:
This will still break Google oauth users though, since we changed from
displayName to ID

dpgaspar:
@mattk42 your right.... give me some time to do some tests and think about a solution to this.

dpgaspar:
Hi @mattk42 

I'm going to close this, I've done it the way you proposed on commit 7be8a38. This is now on release 1.9.0

Thank you!

PR #474, opened 2017-06-05
Fix missing menu dividers

mistercrunch:
Menu dividers have stopped working for me a while back. I'm not sure what broke it, but this fixes them.


dpgaspar:
Nice! thks.

PR #483, opened 2017-06-12
added ldap username format

AbdullahDahmash:
Allow the user to add new option which is AUTH_LDAP_USERNAME_FORMAT

For example:
adding the following line to the configuration:

`````python
AUTH_LDAP_USERNAME_FORMAT = "System1-%s"
`````

When a user as 'flask-user', fab convert it to `System1-flask-user` to login to the ldap server


dpgaspar:
Hi @AbdullahDahmash,

This is a simple and clean commit, thanks. But could you please be more specific about the use case, i'm not very familiar with 'flask-user'.

AbdullahDahmash:
Thanks for replying @dpgaspar 
Currently, I'm experiencing a problem with flask-builder because ldap server (openldap) requires the username string in special format.  
There are actually two usecases: 
Example we have a user with username "dpgaspar". (flask-user is just an example)
1. the users of the system in ldap have initials such as "usr-", when it login in the system, it converts it to "usr-dpgaspar". so we use 
`````
AUTH_LDAP_USERNAME_FORMAT = "usr-%s"
`````
2. openldap requires whole string as username such as "uid=dpgaspar,ou=people,dc=github,dc=org", so we use
`````
AUTH_LDAP_USERNAME_FORMAT = "uid=%s,ou=people,dc=github,dc=org"
`````

I hope it is clear :)

dpgaspar:
Yes, it's clear. But you're number 2 got me worried, this shouldn't be needed.

did you set: AUTH_LDAP_SEARCH and AUTH_LDAP_UID_FIELD ?

PS:
flask-user is actually an open source project :)

AbdullahDahmash:
My mistake about the flask-user :)

The openldap which I use doesn't have indirect user and it doesn't accept usernames without the formatting which i mentioned in the previous comment.  Also, it is not a new user use case. 
I forked this project and test it and it worked

dpgaspar:
I believe it works. My goal here is too help. 

The indirect bind (indirect user) is a "service" user, or a user you agree on with your DevOps/SysAdmins/SecOps or who is responsible for the OpenLDAP system.
It's a standard procedure to first bind using this user, and then search and bind with the login credentials from your app user.

That being said, it would be great if you add some docs for this new config var: https://github.com/dpgaspar/Flask-AppBuilder/blob/master/docs/config.rst

Thank you @AbdullahDahmash

AbdullahDahmash:
I love following standards but sometime it is maybe not possible in some cases :)

Thank you for your time and effort maintaining this repo

dpgaspar:
I understand completely. Thank you for your contribution.

PR #484, opened 2017-06-13
change the indent

gzponline:
sorry I am writing a login_db.html, and find some indent that make me puzzled ,  I think your code is beautifull and conciseness so do some maybe useless change to make others understand the  structure of html


dpgaspar:
It's better now, thks!

PR #491, opened 2017-06-14
[api/update] only update keys specified in POST data

mistercrunch:
Also added basic unit test coverage for read, create, and update endpoint of the API


dpgaspar:
Hi @mistercrunch,

Thank you, can you submit a version compliant with python 2.6 also?

dpgaspar:
Great!

PR #503, opened 2017-06-17
FileUploadField: process_on_store() and process_on_delete()

jak137:
I propose to add two methods to FileUploadField, which can be overridden to customize file uploading.


dpgaspar:
Thank you Jarosław.

PR #511, opened 2017-06-19
fix babel-extract command

RoganW:
support `-k gettext` parameter when extract translatable strings using `fabmanager babel-extract`.

### DIFF
--- a/docs/fabmanager.rst
+++ b/docs/fabmanager.rst
@@ -52,6 +52,15 @@ Command Line uses the excellent click package, so you can have a detailed help f
     --help                          Show this message and exit.
 
 
+**babel-extract** - Babel, Extracts and updates all messages.
+----------------------------------------
+
+Use multi **-k** options separated by space to specify how to locate the strings you want to translate. 
+Default values: **lazy_gettext, gettext, _, __**.
+For example:
+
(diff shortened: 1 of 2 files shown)


dpgaspar:
Sorry @RoganW,

I made a simple test and babel-extract included my test text in translations file (messages.po).

I added a simple gettext to the views.py

Could you clarify?

RoganW:
@dpgaspar There was a problem when use `import gettext as __`
The test steps:
* add test code


* extract translatable strings by `fabmanager babel-extract`
`fabmanager babel-extract --target flask_appbuilder/translations/`

* open babel/messages.pot
There added 3 new translation records:


But there are no translation record for  `"Test gettext()"`

dpgaspar:
Ok, thanks for the reply. I think that the best would be to add a new optional parameter that accepts multiple values, take a look at: http://click.pocoo.org/5/options/#tuples-as-multi-value-options

This way we could add whatever exception(s) one would like.

And then please add some docs to support it: https://github.com/dpgaspar/Flask-AppBuilder/blob/master/docs/fabmanager.rst

Thank you!

dpgaspar:
I would like to release new 1.9.1 tomorrow, will you be able to it until then?

RoganW:
@dpgaspar Done.

dpgaspar:
Great! thank you.

PR #528, opened 2017-06-28
support for Enum type (sqla only)

### LINKED ISSUE #408: Enum Columns is not displaying the right widget
I am trying to use the Enum sqlalchemy type and I was expecting to get a combobox with all valid options. However, that was not the case. Is this a bug, a feature which is not implemented or I am doing something wrong?

jak137:
This is an attempt to support Sqlalchemy Enum types. I don't use Mongo so I'm not sure if there is an equivalent type there. Should resolve #408, I guess.

I also used destructuring-bind for a loop over conversion_table to increase readability.

(tested with Python 3.4)

### DIFF
--- a/flask_appbuilder/fields.py
+++ b/flask_appbuilder/fields.py
@@ -4,7 +4,7 @@ import operator
 
 from wtforms import widgets
 from wtforms.compat import text_type, string_types
-from wtforms.fields import SelectFieldBase, Field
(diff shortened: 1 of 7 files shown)


dpgaspar:
Hi Jarosław,

This is a great feature, give me some time to run some tests.

dpgaspar:
Hi Jarosław,

Nice work, thanks!

PR #537, opened 2017-07-05
Update quickcharts.rst

quanpower:
change "from flask_appbuilder.model.sqla.interface import SQLAInterface" to "from flask_appbuilder.models.sqla.interface import SQLAInterface"


dpgaspar:
Right! thanks

PR #539, opened 2017-07-06
Update quickfiles.rst

quanpower:
change "from flask_appbuilder.model.mixins import ImageColumn" to "from flask_appbuilder.models.mixins import ImageColumn"


dpgaspar:
Thanks!

Issue #544, opened 2017-07-11
SQL injection 


JherekCorn:
allowing string concatenation instead of parameterize-ing input

when using superset:  Interface.py line 70 the call
```
query = query.order_by(order_column + ' ' + order_direction)
```
https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/models/sqla/interface.py#L70

when passing a single quote
```
http://127.0.0.1/dashboardmodelview/list/?_oc_DashboardModelView='
```
The following stack trace occurs
```
Sorry, something went wrong

500 - Internal Server Error

Stacktrace

         Traceback (most recent call last):
  File "/opt/superset/.env/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/superset/.env/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/superset/.env/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/superset/.env/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/superset/.env/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/opt/superset/.env/lib/python2.7/site-packages/flask_appbuilder/security/decorators.py", line 26, in wraps
    return f(self, *args, **kwargs)
  File "/opt/superset/.env/lib/python2.7/site-packages/flask_appbuilder/views.py", line 450, in list
    widgets = self._list()
  File "/opt/superset/.env/lib/python2.7/site-packages/flask_appbuilder/baseviews.py", line 870, in _list
    page_size=page_size)
  File "/opt/superset/.env/lib/python2.7/site-packages/flask_appbuilder/baseviews.py", line 786, in _get_list_widget
    count, lst = self.datamodel.query(joined_filters, order_column, order_direction, page=page, page_size=page_size)
  File "/opt/superset/.env/lib/python2.7/site-packages/flask_appbuilder/models/sqla/interface.py", line 103, in query
    order_direction=order_direction)
[...]

PR #549, opened 2017-07-12
#544 fix for possible sql injection on order by clauses

dpgaspar:


JherekCorn:
The testing looks good, thank you for your great support of a great product!

PR #550, opened 2017-07-13
check whether `session_form_edit_pk` still exist in db

RoganW:
It will raise "AttributeError: 'NoneType' object has no attribute 'id' ", when session_form_edit_pk doesn't exist in database.


dpgaspar:
Thank you!

PR #553, opened 2017-07-15
fix for AttributeError when edit_columns on a view in related_views does not include relationship

jak137:
I am not sure if this is the right place to address this issue but it worked for me...

The problem is that if one has two related models and wants to disable changing relationship then editing via related view raises "AttributeError: 'DynamicForm' object has no attribute ...".

An example:

```
class X(Model):
    id = Column(Integer, primary_key=True)
    name = Column(String, nullable=False)
    note = Column(String)
    def __repr__(self): return self.name

class Y(Model):
    id = Column(Integer, primary_key=True)
    name = Column(String, nullable=False)
    note = Column(String)
    x_id = Column(Integer, ForeignKey(X.id), nullable=False)
    x = relationship(X)
    def __repr__(self): return self.name
```

```
from .models import X, Y

class YView(ModelView):
    datamodel = SQLAInterface(Y)
    add_columns = ['x', 'name', 'note']
    list_columns = ['x', 'name', 'note']
    show_columns = ['x', 'name', 'note']

    # will raise AttributeError: 'DynamicForm' object has no attribute 'x'
    # if one tries:
    # XY -> X -> (pick one) Show -> List Y (inside Show X) -> (pick one) Edit -> change note -> Save
    # the following works as expected:
    # XY -> Y -> (pick one) Edit -> change note -> Save
    edit_columns = ['note']

    # this works fine in both cases, but changing relationship can not be prevented in:
    # XY -> Y -> (pick one) Edit
    # edit_columns = ['x', 'note']

class XView(ModelView):
    datamodel = SQLAInterface(X)
    related_views = [YView]
    add_columns = ['name', 'note']
    list_columns = ['name', 'note']
    show_columns = ['name', 'note']
    edit_columns = ['note']

appbuilder.add_view(XView, "X", category="XY")
appbuilder.add_view(YView, "Y", category="XY")
```

### DIFF
--- a/flask_appbuilder/baseviews.py
+++ b/flask_appbuilder/baseviews.py
@@ -1011,8 +1011,9 @@ class BaseCRUDView(BaseModelView):
         for filter_key in exclude_cols:
             filter_value = self._filters.get_filter_value(filter_key)
             rel_obj = self.datamodel.get_related_obj(filter_key, filter_value)
-            field = getattr(form, filter_key)
-            field.data = rel_obj
+            if hasattr(form, filter_key):
+                field = getattr(form, filter_key)
+                field.data = rel_obj
 
     def pre_update(self, item):
         """


dpgaspar:
Good point.

PR #555, opened 2017-07-17
Allow --app to come from ENV

amancevice:
If `FAB_APP` is defined it overrides the default 'app' value for the `--app` option in `fabmanager`.
Order of preference is `User input > $FAB_NAME > default 'app'`


amancevice:
this looks abandoned. closing

PR #597, opened 2017-08-23
Release/v1.9.4

dpgaspar:
New version 1.9.4 with font-awesome update


mistercrunch:
Thanks for the release! I just bumped Superset to use it and it should be in production at Airbnb today.

PR #618, opened 2017-09-29
Allow dynamic user info resolution for OAuth2

codemug:
The code contains user info extraction handles for well known OAuth2 providers in `security.manager.AbstractSecurityManager.get_oauth_user_info` but what about the other, less known providers? Its a way of saying that FAB only supports Google, Linkedin, GitHub and Twitter for OAuth2 authentication, which somewhat defeats the purpose of a standard protocol for authentication.

What the PR does:
Adds the ability to be able to authenticate against ANY OAuth2 provider as long as it exposes the user info.

How:
It adds two extra params to the OAuth2 provider configuration, namely:
`user_query`: The query to get user info object. This is appended to the basepath.
`user_resolver`: A resolver method that parses the fetched user info object into a FAB user model.


dpgaspar:
Thank you, please give me some time to test this.

codemug:
No problem, take your time :)

dpgaspar:
I don't see the point if someone can use the decorator for OAuth:

````
    @appbuilder.sm.oauth_user_info_getter
    def my_user_info_getter(sm, provider, response=None):
        if provider == 'github':
            me = sm.oauth_remotes[provider].get('user')
            return {'username': me.data.get('login')}
        else:
            return {}
````

You can implement whatever you want right?

codemug:
Right, I didn't know about the decorator, which is definitely a much more cleaner way to handle this.

Kerneels:
Hi. Please provide more info on how exactly to use the decorator, in particular, where should this function actually go, and how would one know what to import for it to work? Thanks!

PR #625, opened 2017-10-11
Make updating perms on init optional

mistercrunch:
Whenever we re-deploy Superset at Airbnb, we restart 5 web servers which
all have a set of many Gunicorn workers.

Each one of these workers is a process and inits the FAB app, resulting
in a fair amount of database contention around perms tables, and sometimes ultimately
leading to bad database state as I suspect that the perms merge
operations aren't well transaction-insulated. We've added unique
constraints on the DB preventing that, but that's not a great solution.

This PR adds a flag to AppBuilder allowing to turn off perm updates on
init.

The default behavior stays unchanged though.

Note that I decided to add the condition inside `_add_menu_permissions` and `_add_permissions` as opposed to adding them in the calling methods since they `_add_permissions` is called multiple times and could be called in more places in the future. Adding the flag where perms are applied seemed less brittle. Happy to change it though if needed.

### DIFF
--- a/flask_appbuilder/base.py
+++ b/flask_appbuilder/base.py
@@ -97,7 +97,8 @@ class AppBuilder(object):
                  base_template='appbuilder/baselayout.html',
                  static_folder='static/appbuilder',
                  static_url_path='/appbuilder',
-                 security_manager_class=None):
+                 security_manager_class=None,
+                 update_perms=True):
         """
             AppBuilder constructor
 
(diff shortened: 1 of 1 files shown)


timifasubaa:
LGTM

dpgaspar:
Sorry for the delay, merge and release it ASAP

dpgaspar:
This is out with v1.9.5

PR #626, opened 2017-10-12
Add translation for zh_HK - Traditional Chinese 

kelvinliuhk:
Please add this translation. and the naming should update like that. Thanks.

zh_HK - Traditional Chinese
zh - Simplified Chinese

LANGUAGES = {
    'zh_HK': {'flag':'hk', 'name':'Traditional Chinese'},
    'zh': {'flag':'cn', 'name':'Simplified Chinese'}
}


dpgaspar:
Great, thank you for this!

dpgaspar:
This is out with v1.9.5

PR #628, opened 2017-10-14
Remove double-instantiation of declarative base

harmon:
The Model is already decorated with @as_declarative, so when SQLA wraps that again in a call to declarative_base(), it's creating yet another instance of declarative_base. This is causing flask_migrate (which uses SQLA.Model.metadata) to not recognize all the tables I've declared when I subclass from Model (since the metadata instances are different).


dpgaspar:
Release on 1.11.0, Thank you!

PR #631, opened 2017-10-20
Add binary column support for sqlalchemy

jgao54:
Binary columns are not currently supported in FAB. Models with Binary columns cannot be rendered properly when using the default modelview.


criccomini:
@mistercrunch can you look at this please?

dpgaspar:
Thanks!

dpgaspar:
This is out with v1.9.5

PR #639, opened 2017-10-31
Add composite pk support for sqlalchemy

jgao54:
Add composite primary key support for SQLAlchemy

### DIFF
--- a/flask_appbuilder/models/base.py
+++ b/flask_appbuilder/models/base.py
@@ -213,6 +213,9 @@ class BaseInterface(object):
     def is_pk(self, col_name):
         return False
 
+    def is_pk_composite(self):
+        raise False
+
     def is_fk(self, col_name):
         return False
 
@@ -254,7 +257,10 @@ class BaseInterface(object):
             return a list of pk values from object list
         """
         pk_name = self.get_pk_name()
-        return [getattr(item, pk_name) for item in lst]
+        if self.is_pk_composite():
+            return [[getattr(item, pk) for pk in pk_name] for item in lst]
+        else:
+            return [getattr(item, pk_name) for item in lst]
 
     def get_pk_name(self):
         """
@@ -263,7 +269,11 @@ class BaseInterface(object):
         raise NotImplementedError
 
     def get_pk_value(self, item):
-        return getattr(item, self.get_pk_name())
+        pk_name = self.get_pk_name()
+        if self.is_pk_composite():
+            return [getattr(item, pk) for pk in pk_name]
+        else:
+            return getattr(item, pk_name)
 
     def get(self, pk, filter=None):
         """
--- a/flask_appbuilder/models/sqla/interface.py
+++ b/flask_appbuilder/models/sqla/interface.py
@@ -271,6 +271,9 @@ class SQLAInterface(BaseInterface):
         except:
             return False
 
+    def is_pk_composite(self):
+         return len(self.obj.__mapper__.primary_key) > 1
+
     def is_fk(self, col_name):
         try:
             return self.list_columns[col_name].foreign_keys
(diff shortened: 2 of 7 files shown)


jgao54:
@mistercrunch  @dpgaspar 
Please take a look when you get a chance. I'm not very familiar with the full layout of FAB, so let me know if there's any issues with this implementation and I can make adjustment as necessary!

flask_appbuilder/models/sqla/interface.py
```diff
@@ -516,9 +516,11 @@ def get(self, id, filters=None):
         return self.session.query(self.obj).get(id)
 
     def get_pk_name(self):
-        for col_name in self.list_columns.keys():
-            if self.is_pk(col_name):
-                return col_name
+        pks = [pk.name for pk in self.obj.__mapper__.primary_key]
```
mistercrunch: The name `get_pk_name` is becoming a bit misleading as it may contain a name or a list. I'm not big on methods that return heterogenous types.

I'd go with renaming it `get_pk` and make sure it's always a list and handled as a list by the callers.

flask_appbuilder/views.py
```diff
@@ -570,12 +570,26 @@ def action_post(self):
...
+    def _eval_pk(self, pk):
+        """
+        In the case of a model with composite primary keys, `rowid` is a string
+        representation of a list. This method checks for this case and converts
+        it to a list if necessary.
+        """
+        if pk.startswith('[') and pk.endswith(']'):
+            # handle case where composite pk contains datetime column
+            import datetime
+            return eval(pk)
```
mistercrunch: evals are bad, why is rowid a string rep or a list in the first place? Can we change that instead?

flask_appbuilder/models/base.py
```diff
@@ -254,7 +254,10 @@ def get_keys(self, lst):
             return a list of pk values from object list
         """
         pk_name = self.get_pk_name()
-        return [getattr(item, pk_name) for item in lst]
+        if isinstance(pk_name, str):
```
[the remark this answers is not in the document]
jgao54: ah good point. Thanks!

flask_appbuilder/views.py
jgao54: @mistercrunch let me see if I can get around this. This was a fairly quick solution using the existing setup, but i agree eval isn't ideal, I'll dig a bit more into it.

flask_appbuilder/models/sqla/interface.py
jgao54: 👍  cool, i originally kept it as str to minimize breakage (by keeping backward compatibility), but maybe a more consistent return type is better

jgao54:
@mistercrunch  thanks for the feedback, let me dig a bit more into the framework and make it more consistent. You actually addressed all my uncertainties with this PR!

jgao54:
@mistercrunch 
I had another look at the code. When an action is performed in a ModelView (i.e. multi-delete), each row selection is identified by their input value, which is the primary key. Even when the value is a list in the case of composite PKs, it gets converted to a string presentation in the html:

```{% if actions %}
    <td>
        <input id="{{pk}}" class="action_check" name="rowid" value="{{pk}}" type="checkbox">
    </td>
{%  endif %}```

Though SQLAlchemy supports querying by composite PKs as a list, but html value is text only afaik.

mistercrunch:
JSON serilalize the pk? @dpgaspar what do you think?

jgao54:
@dpgaspar would be awesome if you can take a look at this one. We are in the process of migrating Airflow to Flask-Appbuilder, and this is a major blocker for us. Thank you!

dpgaspar:
I'm just looking at this now, created a simple example:
models.py 
````
# models.py
class Invoice(Model):
    invoice = Column(Integer, primary_key=True)
    ref_num = Column(Integer, primary_key=True)
    name = Column(String(50))

    def __repr__(self):
        return self.name

# views.py
class InvoiceModelView(ModelView):
    datamodel = SQLAInterface(Invoice)
    add_columns = ['invoice', 'ref_num', 'name']
    list_columns = ['invoice', 'ref_num', 'name']
````

You get the point, but on list->show record fails with "TypeError: hasattr(): attribute name must be string"

jgao54:
@dpgaspar  ah good catch! will work on this

dpgaspar:
@jgao54 great thks!

jgao54:
@dpgaspar  please review again! Thanks. Hopefully this fixes the issues.
The get/update/delete api are not supporting composite PK right now, but want to save that for another PR.

mistercrunch:
@jgao54 please add test coverage on this

jgao54:
@mistercrunch  added tests.  I'm pretty concerned with the `eval` as well, looking into code to serialize pk

jgao54:
@mistercrunch  converted to serialize/deserialize composite pk.

mistercrunch:
LGTM at this point, will let @dpgaspar click the green button.

dpgaspar:
Looks good, clicking the green button. Great PR @jgao54, i'll add and example and some docs

PR #648, opened 2017-11-18
Redirects on invalid list view page number

### LINKED ISSUE #637: Pagination: Changing page_size doesn't update the page number
Forwarding issue from [apache/incubator-superset](https://github.com/apache/incubator-superset)
apache/incubator-superset#3731

This is due to missing logic in the handling of pagination.
Changing the page_size changes the number of total pages, so there are cases where a page number no longer exist in the pagination with the new page_size.

I would open a PR but I don't know what's the best way to handle this. I have to proposal:
- redirect the user on the last page available in case of page "overflow"
- silently show the last page even for page numbers exceeding the total page number.

baldoalessandro:
Instead of returning an empty set redirect to the last valid page
given the provided pagination parameters.

Fixes #637

### DIFF
--- a/flask_appbuilder/baseviews.py
+++ b/flask_appbuilder/baseviews.py
@@ -789,6 +789,13 @@ class BaseCRUDView(BaseModelView):
             order_column, order_direction = self.base_order
         joined_filters = filters.get_joined_filters(self._base_filters)
         count, lst = self.datamodel.query(joined_filters, order_column, order_direction, page=page, page_size=page_size)
+
+        if page:
+            page_number = count // page_size
+            if page > page_number:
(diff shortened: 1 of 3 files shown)


dpgaspar:
This fails on python 2.6, can you fix it?

baldoalessandro:
Yes I can, but I was hoping for your advice/ review on the proposed
solution. Probably I cannot work on this until the next week though.

dpgaspar:
I've tested it and looks good, and the solution is simple. The trick using the exception ValueError and redirecting to it's message is a bit weird thought, but simple no doubt.

( maybe creating a custom Exception for this, would be a gain for readability)

PR #655, opened 2017-11-29
Add feature to allow form prefill

jgao54:
This is a feature currently in flask_admin that I find to be pretty useful.

When an edit view is rendered, sometimes we want to preprocess/prefill the form when we have extra fields available that are not part of the model but depends on the data in the model. This makes it possible to render the processed edit view.

### DIFF
--- a/flask_appbuilder/baseviews.py
+++ b/flask_appbuilder/baseviews.py
@@ -970,6 +970,9 @@ class BaseCRUDView(BaseModelView):
         else:
             # Only force form refresh for select cascade events
             form = self.edit_form.refresh(obj=item)
+            # Perform additional actions to pre-fill the edit form.
+            self.prefill_form(form, pk)
+
         widgets = self._get_edit_widget(form=form, exclude_cols=exclude_cols)
         widgets = self._get_related_views_widgets(item, filters={},
                                                   orders=orders, pages=pages, page_sizes=page_sizes, widgets=widgets)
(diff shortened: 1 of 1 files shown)


jgao54:
@mistercrunch could you take a look at this one too?

mistercrunch:
Can you add a simple example in the docstring of how to fill in a text field for instance, otherwise LGTM

mistercrunch:
Random related comment - the Django Admin has this cool feature where you can pass default form field values as GET params.

jgao54:
Example added!

I just looked up the Django Admin feature mentioned above, it's is slightly more complicated to implement and doesn't provide as much flexibility as passing the form reference, good to know though!

flask_appbuilder/baseviews.py
```diff
@@ -1015,6 +1018,23 @@ def _fill_form_exclude_cols(self, exclude_cols, form):
...
+        """
+            Override this, will be called only if the current action is rendering
+            an edit form, and is used to perform additional action to
+            prefill the form.
+
+            This is useful when you have added custom fields that depend on the
+            database contents. Fields that were added by name of a normal column
+            or relationship should work out of the box.
+
+            Example:
```
mistercrunch: for RST to render the code block to render properly I think you need two colons as in `example::`
https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/baseviews.py#L358

jgao54: gotcha, done.

dpgaspar:
Thank you @jgao54 and @mistercrunch, this is out with v1.9.5

PR #656, opened 2017-12-01
Modify the sample code in the document

braveghz:


dpgaspar:
Thank you @braveghz,

Just a couple of comments on this: 
Don't really agree with the comments indentation, my version isn't correct also, we should use something like:

```python

    @expose('/method1/<string:param1>')
    def method1(self, param1):
        """
        do something with param1
        and return to previous page or index
        """
        return 'Hello %s' % (param1)
      
```
Also further on the docs, when explained the use of @has_access the example is not coherent, you could change it also.

PR #661, opened 2017-12-07
Add feature to allow form to be processed prior to populating model

jgao54:
This allows the form to be processed upon submission. 

Useful if there are extra fields submitted that are not part of the data model attributes, and need to be processed prior to populating the data model.

cc  @mistercrunch


mistercrunch:
🥇 🚀

PR #663, opened 2017-12-08
Support AUTH_USER_REGISTRATION with AUTH_REMOTE_USER

ottomata:
I'm using AUTH_REMOTE_USER with LDAP auth done by an HTTP reverse proxy.  The [docs say](http://flask-appbuilder.readthedocs.io/en/latest/security.html#authentication-methods):

> Finally you can allow users to self register...These settings can apply to all the authentication methods

which is not true.  This patch makes it true for AUTH_REMOTE_USER.

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -688,12 +688,28 @@ class BaseSecurityManager(AbstractSecurityManager):
             :type self: User model
         """
         user = self.find_user(username=username)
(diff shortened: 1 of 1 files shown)


dpgaspar:
Right you are, thks!

ottomata:
Great!  I'd love to get this updated in superset asap, what's the timeline for a release bump?

elukey:
ping :)

PR #667, opened 2017-12-10
Docs and example for composite keys

dpgaspar:
Simple example to show the use of composite keys, and their reference with SQLA realationships.

Follow up on merged PR #639


jgao54:
Oh this is great, thanks @dpgaspar

PR #678, opened 2017-12-21
Implement create-user command to fabmanager

wrp:
It is convenient to have a scriptable method for creating users.  It might be cleaner to merge this with 'create-admin', but I don't see an easy way to do that without either changing the name of 'create-admin', or having usage 'fabmanager create-admin --role=non-admin role', which just seems wrong.

PR #686, opened 2018-01-09
Remove Travis support for Python 2.6 since SQLAlchemy does not sup…

MichaelPereira:
…port it anymore

This will unblock all other PR since they are affected by this issue (like #678)


dpgaspar:
Agree, thank you once more @MichaelPereira

PR #694, opened 2018-01-26
Accept SQLAlchemy custom types defined by TypeDecorator

jgao54:
This will allow any [custom sa types](http://docs.sqlalchemy.org/en/latest/core/custom_types.html) (e.g. UtcDateTime in Airflow) inherited from TypeDecorator class. 

cc @mistercrunch

### DIFF
--- a/flask_appbuilder/models/sqla/interface.py
+++ b/flask_appbuilder/models/sqla/interface.py
@@ -25,6 +25,10 @@ def _include_filters(obj):
         if not hasattr(obj, key):
             setattr(obj, key, getattr(filters, key))
 
+def _is_sqla_type(obj, sa_type):
+    return isinstance(obj, sa_type) or \
+        isinstance(obj, sa.types.TypeDecorator) and isinstance(obj.impl, sa_type)
+
 
 class SQLAInterface(BaseInterface):
     """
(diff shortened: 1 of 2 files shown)


flask_appbuilder/models/sqla/interface.py
```diff
@@ -158,61 +162,61 @@ def is_file(self, col_name):
 
     def is_string(self, col_name):
         try:
-            return isinstance(self.list_columns[col_name].type, sa.types.String)
+            return _is_sa_type(self.list_columns[col_name].type, sa.types.String)
```
mistercrunch: I'm confused at the code *before* this change, why to we expect `isinstance` to fail? If it's really a thing, maybe `_is_sa_type` should handle it though.

flask_appbuilder/models/sqla/interface.py
```diff
@@ -25,6 +25,10 @@ def _include_filters(obj):
         if not hasattr(obj, key):
             setattr(obj, key, getattr(filters, key))
 
+def _is_sa_type(obj, sa_type):
```
mistercrunch: [optional] `_is_sqla_type` is more clear than `_is_sa_type` to me

flask_appbuilder/tests/test_sqlalchemy.py
```diff
@@ -0,0 +1,22 @@
+from nose.tools import eq_
+import unittest
+import sqlalchemy as sa
+from flask_appbuilder.models.sqla.interface import _is_sa_type
+
+class CustomSqlaType(sa.types.TypeDecorator):
+		impl = sa.types.DateTime(timezone=True)
```
mistercrunch: indentation seems off

flask_appbuilder/tests/test_sqlalchemy.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class NotSqlaType():
+		def __init__(self):
```
mistercrunch: indentation seems off

flask_appbuilder/tests/test_sqlalchemy.py
```diff
@@ -0,0 +1,22 @@
...
+			self.impl = sa.types.DateTime(timezone=True)
+
+
+class FlaskTestCase(unittest.TestCase):
+	def test_is_sa_type(self):
```
mistercrunch: are these spaces tabs?

flask_appbuilder/models/sqla/interface.py
jgao54: sounds good

flask_appbuilder/models/sqla/interface.py
jgao54: I believe these type checks are used as process of elimination to map model fields to WTForm fields.  Each model field iterates through all the types checks until it finds the one that returns True, and then do the conversion accordingly ([ref here](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/forms.py#L37-L50)).

So if the type didn't exist, then the field doesn't get processed and that breaks the view page.

mistercrunch:
LGTM expect for failing tests.

jgao54:
fixed, thanks!

flask_appbuilder/models/sqla/interface.py
dpgaspar: It's a fail safe for self.list_columns[col_name].type

dpgaspar:
I like it, it's simple

jgao54:
Just discovered that sorting/filtering on custom type isn't supported yet. Got the following error message when I attempted to filter a field with custom type:

```StatementError: (exceptions.TypeError) expected datetime.datetime, not u'2018-02-06 01:36:57' [SQL: u'SELECT count(%s) AS count_1 \nFROM log \nWHERE log.execution_date > %s'] [parameters: [{}]]```

Wil look into adding support for this.

Issue #699, opened 2018-02-03
Non-unique associations


john-bodley:
On Superset we've noticed instances where the associations defined by the following tables:
- ab_permission_view
- ab_permission_view_role
- ab_user_role

are non-unique, i.e., queries of the form
```
SELECT 
    user_id,
    role_id,
    COUNT(1)
FROM 
    ab_user_role
GROUP BY 
    user_id,
    role_id
HAVING COUNT(1) > 1
```
return a number of records. Although I haven't seen this cause a problem in the web app REST API calls for updating the users associated with a specific role (`/roles/api/update/<id>`) will fail. It seems the solution is simply to enforce uniqueness constraints on these tables.

dpgaspar:
Release on 1.9.7

PR #700, opened 2018-02-03
[constraints] Adding uniqueness constraints

john-bodley:
This PR addresses issue https://github.com/dpgaspar/Flask-AppBuilder/issues/699. 

cc: @mistercrunch


mistercrunch:
I've been adding those manually at the physical database level before. This is the right place for them.

PR #712, opened 2018-02-17
Use `dateutil.parser.parse` method instead of `strptime` 

jgao54:
datetime.strptime doesn't handle the parsing of timezone-aware dates very well; plus this is easier to read.


jgao54:
friendly ping @mistercrunch  @dpgaspar

PR #716, opened 2018-02-21
Update baseviews.py

cw1427:
Fix a bug when blueprint did not pass a root_path, it will use the import_name as the root_path to search the template.  But the __name__ is the parent class BaseView's import_name, so for the sub-classes which will not get its correct root path when searching the template file in the template path. So change it to the __module__ which can identify its correct root path for the blueprint template path.

### DIFF
--- a/flask_appbuilder/baseviews.py
+++ b/flask_appbuilder/baseviews.py
@@ -126,11 +126,11 @@ class BaseView(object):
         self.static_folder = static_folder
         if not static_folder:
             # Create blueprint and register rules
-            self.blueprint = Blueprint(self.endpoint, __name__,
+            self.blueprint = Blueprint(self.endpoint, self.__module__,
                                        url_prefix=self.route_base,
                                        template_folder=self.template_folder)
         else:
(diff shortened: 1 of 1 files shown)


cw1427:
Hi guys, Please take a look at this PR which want to fix a bug when use different template path for different blueprint. Especially for the addon blueprint's template path which should be in the addon folder but not the App root. So I push this PR for fix.

dpgaspar:
Sorry for the delay, I'll test this ASAP

flask_appbuilder/baseviews.py
```diff
@@ -64,7 +64,9 @@ class BaseView(object):
     """ The template folder relative location """
     static_folder = 'static'
     """  The static folder relative location """
+    static_url_path = '/static'
     base_permissions = None
+    base_permissions_inclusive = ['menu_access']
```
dpgaspar: Can you explain the purpose of setting a default 'menu access' on ModelView?

dpgaspar:
Left some comments, and this to be rebased

PR #757, opened 2018-06-05
HOTFIX-754 LDAP authentication requires too permissive LDAP ACLs

LLluma:
It is looking like duct tape but it works. I would deeply refactor LDAP part of BaseSecurityManager.


LLluma:
Hm... It is looking 'develop' branch is not synced with 'master'. I'll create pull request to master.

PR #758, opened 2018-06-05
HOTFIX-754 LDAP authentication requires too permissive LDAP ACLs

LLluma:
It is looking like a duct tape but it works. I would deeply refactor LDAP part of BaseSecurityManager.


dpgaspar:
Looks good!

Issue #759, opened 2018-06-08
Select all checkbox in related view selects checkboxes from all related views


nachtkinder:
# Steps to reproduce

1. Create a model (A) that has a relationship with at least 2 other models (B, C)
1. Create views for all three models
1. Specify within view for A that views B and C are related (related_views = [B, C])
1. Create a new object using model A (oA) that has objects created from both B (oB) and C (oC) related to it
1. Browse to the show/edit page for oA
1. Switch to the tab listing related oB objects
1. Use the select all checkbox to select all rows in the table
1. Switch to the tab listing related oC objects
1. Note that all oC objects are selected also.

# Cause

The crux of this issue is that the jQuery selector used to identify which checkboxes should be affected by the check_all option is not specific enough to exclude checkboxes from other related views.

I'll issue a pull request with a fix related to this issue number.

Regards,
Nachtkinder

dolfandringa:
HI @nachtkinder Your code looks good, but I would like to verify the problem and the fix both. I just made a small skeleton app, but probably missed a step somewhere. In Flask Appbuilder on the list page (step 6/7) I don't have checkboxes allowing me to select all rows. What kind of relationship are you using? Do you have code to a demo app? Or suggest a change to [this single page one](https://gist.github.com/dolfandringa/1d8f837235f2d841d0dcfc3d1dbc1f81) that shows the problem?

nachtkinder:
Sorry @dolfandringa,

In the steps to reproduce I neglected to mention that you need to define an action for the related views.

Add action to the imports:

```
from flask_appbuilder.actions import action
```

And to each of CourseView and ProgramView add:

```
@action("muldelete", "Delete", "Delete all Really?", "fa-rocket", single=False)
def muldelete(self, items):
    self.datamodel.delete_all(items)
    self.update_redirect()
    return redirect(self.get_redirect())
``` 

Then when you edit a student you should see the checkboxes.

Regards,
Nachtkinder

PR #760, opened 2018-06-08
Fix for issue #759

nachtkinder:
This is a fix for issue #759 relating to checkboxes in related views.


dpgaspar:
Hi @nachtkinder,

Yes this is a bug, but this solution does not seem solve it, using @dolfandringa example, the action bottom deletes the wrong record (Show student, use action to delete Program, deletes the course and we get an internal server error)

PR #763, opened 2018-06-15
Add french flag

emirot:


emirot:
@dpgaspar should it be added or closed ?

PR #770, opened 2018-06-21
Fix indirect bind when using append/format

ledor473:
We are running into an issue when using both `AUTH_LDAP_APPEND_DOMAIN` and indirect bind (` AUTH_LDAP_BIND_USER` + `AUTH_LDAP_BIND_PASSWORD`).

Here's the debug logs that we used to discover the issue:
```
LDAP indirect bind with: CN=indirect-bind0user,OU=company,DC=com
LDAP BIND indirect OK
LDAP got User [('CN=my-username,OU=Users,OU=company,DC=com', {'givenName': [b'MyName'], 'mail': [b'myemail@company.com'], 'sn': [b'MyLastName']})]
LDAP bind with: CN=my-username,OU=Users,OU=company,DC=com XXXXXX
...
And here's the log we would have seen if the `con.bind_s` was successfull 
...
LDAP bind OK: CN=my-username,OU=Users,OU=company,DC=com@company.com

In our case the `_bind_ldap` eventually returns False
```


dpgaspar:
Hi @ledor473,

Thank you for contributing, but I'm confused, if your using AUTH_LDAP_APPEND_DOMAIN and AUTH_LDAP_BIND_USER, why do you want to skip the append domain? just don't use it.

Please post more detailed logs and config to clarify. Thks

dpgaspar:
I understand @ledor473, it's not good, but if you skip the AUTH_LDAP_APPEND_DOMAIN everything works ok (without your patch), can you confirm?

dpgaspar:
right, but you've got:

```
            if indirect_user:
                ...
            else: 
                # Only append/format the username if it's not the DN returned by the LDAP indirect bind
                if self.auth_ldap_username_format:
                    username = self.auth_ldap_username_format % username
                if self.auth_ldap_append_domain:
                    username = username + '@' + self.auth_ldap_append_domain
```

So, I guess the same thing will happen, can you confirm?

ledor473:
Oh I see what you mean!
The code I changed is in `_bind_ldap` which is only used to create a working LDAP connection. That connection will later be used to lookup the username in `_search_ldap`: https://github.com/dpgaspar/Flask-AppBuilder/blob/c4a6202aaa4aa64dc7a93644511a9a4cf95fb01c/flask_appbuilder/security/manager.py#L594-L606

dpgaspar:
ok, thanks!

Really need to refactor LDAP auth...

Issue #447, opened 2017-04-13
Not really an issue, just wondering interest value


GregSilverman:
For a project I've been using Flask Appbuilder with to create a patient registry one of the requirements was to extend the built-in search widget to allow for multiple searches on the same field (e.g., a list of Boolean ANDed diagnoses by patient). To do this, I had to extend the query method in the SQLAinterface to allow for arbitrary numbers of self-joins across a model instance. I will make this available as soon as I clean up the code.

Not sure if you would want to merge into your code base, or if I should just fork this. Thoughts?

dpgaspar:
Not sure i'm following this, you can already search multiple times the same field, they will all be ANDed.

Can you be more detailed?

GregSilverman:
Briefly, without going into too much detail, as an example, if I have the class:

`Diagnosis(Model):
    __tablename__ = 'vw_svc_diagnosis'
    diagnosis_id = Column(String(32), primary_key=True)
    dx_name = Column(String(255))
    mrn = Column(String(255))
    dx_code = Column(String(255))`

And I want to query for the set of patients (by MRN) who have 
`dx_name = 'Diagnosis 1' AND dx_name = 'Diagnosis 2' AND dx_name = 'Diagnosis 3'`, etc., out of the box, this will not work, since 

a) get_filter_args in urtools.py and all the functions in filters.py expect a dictionary and thus will not allow multiple filters of the same type on the same attribute

b) to get the above query, I need to do a JOIN across the Diagnosis class linked by the patient MRN.

In any case, I have circumvented this by appropriate overrides in the methods and modules noted above and also in the query method of the sqla/interface.py module.

I am working on trying to generalize it so that it is not so specific to the application for which I had to do this. I will gladly share the code, once it is ready.

PR #791, opened 2018-07-19
Filter now acknowledges multiple values for single field (fixes #789).

### LINKED ISSUE #789: Relation as Many filter does not work correctly when multiple values selected
# Description

Having filtered a list view on a \*-to-many relationship:

 1. the filter field does not acknowledge values beyond the first
 2. the filtered list is incorrect (sometimes showing more results than it should, sometimes fewer or none)

Both of the above occur despite all the values for the selected field being present in the URL query string.

# Cause

The part of Flask Appbuilder that performs filtering of result sets uses `request.args.get('KEY')` in all instances rather than `request.args.getlist('KEY')`.  In the event that request.args['KEY'] has multiple values, `get()` returns only the first of the values whereas `getlist()` returns all values as a list.

# Proposed solution

Modify Flask Appbuilder's filtering to use `request.args.getlist()` in order to process multiple values for a single filter field.

nachtkinder:
These minor changes implement multiple value filtering on a single search field as described in issue #789


dolfandringa:
Hi @nachtkinder thanks for you pull request. It looks like good code, but to make sure existing code doesn't break, and down the road nobody breaks your fix, I would like to ask you to add some unit tests to make sure both old behavior still works but also the problems from #789 #447  remain fixed. Is there anything I am overlooking @dpgaspar?

nachtkinder:
Hi @dolfandringa thanks for taking the time to review this PR.  I must confess that, whilst capable of writing reasonable code, I have little experience in writing unit tests.  Can you recommend some good source material for a beginner?

Regards,
-- 
Nachtkinder

dolfandringa:
@nachtkinder It is a good next step to learn in this case. Initially it can seem daunting, but once you get the hang of it, I feel it makes writing code actually a lot more fun. This looks like a good article: https://code.tutsplus.com/tutorials/beginning-test-driven-development-in-python--net-30137

Specifically for FAB:
1. We are using nosetests for running our tests. But the easiest way to run the tests (which works across multiple test runners) is just to run `python setup.py test`. You will see it run through all the tests. To make all tests run correctly, you need to `pip install flask_mongoengine pillow` because they are normally not installed. And you need a mongodb running. If you don't, it's no real issue, just expect a number of tests to fail because of it. 
2. A lot of our testing is done by calling endpoints (url's) and checking the result. This is more of a black-box/functional testing. For a bugfix, I usually prefer to go down into the code as much as I can (test the smallest unit possible). So test the actual function where the bug is, not just calling the url/endpoint and checking the result. In this case, I would write tests for the add_filter add_filter_index and apply methods.

Lastly, the tests to write are:
1. A test that fails with the current master branch, illustrating the problem, but will succeed with your fix. That way you test that any changes in the future would make sure that your desired behaviour doesn't change.
2. Test some other scenarios that people can use the code that you are changing, and make sure those don't break with your changes (often more important).

To illustrate: I just made  a PR for testing one-to-one relationships (PR #824). I "fixed" some code to make sure relationships were correctly identified as one-to-one relationships (before they weren't). The FAB code didn't have any tests for this part, so I wrote one for the one-to-one relationship.  But I failed to write code testing that one-to-many and other relationships were still detected correctly as well. Turns out, they weren't and my change broke the code. Thanks to our tests and using Travis automatically for it, I saw a bunch of unintended breaks (pretty cryptic errors initially). When I figured out what was happening, I also added tests to test that all other relationships were detected correctly, fixed my code, and now everything is hunky-dory again (and my PR merged, and I became a FAB contributor, and unicorns are jumping over rainbows) .

nachtkinder:
Sorry Guys,
I am still interested in getting this included but have been snowed under at work and have had little time to devote to it.  Hopefully I will be able to find some time to write the requested tests/docs soon.

Regards,
Nachtkinder

dpgaspar:
Hi @nachtkinder,

Docs and tests would be great!

nachtkinder:
Hi Guys,

Guess none of us spotted until now that the pull request would break filtering the list view by multiple values for a non-relationship field.  Am working to fix this and will let you know when it's ready.

Regards,
Nachtkinder

nachtkinder:
Easy, Stalebot, easy boy.

Still interested in fixing this but still struggling to find time.

Regards,
Nachtkinder

PR #797, opened 2018-08-07
Remove safe filter from template

graceguo-supercat:
We found there are vulnerable to stored XSS in multiple list endpoints within the application. For example, user can inject javascript <script> tag and code into First Name, Description field, and javascript code get executed when anyone loaded the list view.

This PR is to propose a solution that remove `safe` filter in templates. In most of template, we don't want to render as HTML and js, so I think remove `safe` won't affect current good content behavior and just auto-escape those tag characters.


mistercrunch:
This LGTM. @dpgaspar are you ok with merging this?

The `| safe` filter works in the context of a Django-like admin where only trusted people can use CRUD, but in the context of FAB, where CRUD is expected to be used by normal users, I think we need to shut it off by default. 
 
@dpgaspar if you think removing `|safe` is a break of backward compatibility and want to allow admin/users to opt in this fix, we **could** have some ModelView class attribute called `unsafe_mode = False` that people could flip on at their own risk.

@dpgaspar what do you think?

flask_appbuilder/templates/appbuilder/general/lib.html
```diff
@@ -190,16 +190,16 @@
           {% if field.type == 'HiddenField' %}
             {{ field}}
           {% else %}
-            {{begin_sep_label|safe}}
+            {{begin_sep_label}}
```
dpgaspar: This needs the pipe safe

flask_appbuilder/templates/appbuilder/general/lib.html
```diff
@@ -190,16 +190,16 @@
...
                 <label for="{{field.id}}" control-label>
                     {{ field.label.text }}
                     {% if field.flags.required %}
                         <strong style="color: red">&#42;</strong>
                     {% endif %}
                 </label>
-            {{end_sep_label|safe}}
-            {{begin_sep_field|safe}}
-                {{ field(**kwargs)|safe }}
+            {{end_sep_label}}
```
dpgaspar: This needs the pipe safe

flask_appbuilder/templates/appbuilder/general/lib.html
```diff
@@ -209,7 +209,7 @@
             {% endfor %}
             </div>
           {% endif %}
-            {{end_sep_field|safe}}
+            {{end_sep_field}}
```
dpgaspar: This needs the pipe safe

dpgaspar: all `*_sep_*` vars need the pipe safe

graceguo-supercat:
updated per comments. Please take a look again. Thank you!

dpgaspar:
Looks good, let me run some more tests

graceguo-supercat:
Thanks for code review  🎉 🎉 🎉

feng-tao:
are we going to have a release for this fix?

mistercrunch:
@dpgaspar , I'm not sure what your release process consist of beyond the standard Pypi push, but I'd be happy to honor your process and push releases following your guidelines.

PR #809, opened 2018-09-21
Fix oauth user not found. improved to search users in both fields

mopemope:
Fix oauth user not found. improved to search users in both fields.


dpgaspar:
but, It already does

PR #811, opened 2018-09-23
Update flask login

jmcarp:
Flask-login 0.2.11 was released in May 2014. There haven't been many breaking changes since then, and there have been some improvements and bug fixes, so this patch upgrades to >=0.3.0.

cc @mistercrunch

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,6 @@
+[run]
+source = flask_appbuilder
+
+omit =
+  *messages.py*
+  *baseapp.py*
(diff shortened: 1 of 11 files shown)


mistercrunch:
This LGTM, though it will require release notes to make sure people know it may affect their local security managers / projects.

dpgaspar:
Ok, This is a breaking change, I expect to include it on the next release this weekend.

I wonder 2.0.0 or 1.12.0, mmm...

PR #813, opened 2018-09-27
Un-pin dependencies.

jmcarp:
Pinning to exact versions can cause conflicts when importing
flask-appbuilder. Pin to minimum versions or version ranges instead.


jmcarp:
This pulls in some useful updates from @fordguo; sorry I didn't notice your PR earlier! This also drops support for python 3.3, which reached end of life a year ago. Since we're already making breaking changes around flask-login, this seems like a good time to drop python 3.3 as well.

cc @mistercrunch

SaurabhBajaj:
Looks good to me 👍

mistercrunch:
@dpgaspar

dpgaspar:
ok, let me check it

SaurabhBajaj:
@dpgaspar  Can we move this PR forward?

flask_appbuilder/models/sqla/__init__.py
```diff
@@ -30,8 +30,7 @@ class SQLA(SQLAlchemy):
 
         Use it and configure it just like flask_SQLAlchemy
     """
-    def make_declarative_base(self, metadata=None):
-        """Creates the declarative base."""
+    def make_declarative_base(self, model, metadata=None):
```
dpgaspar: Can you explain this change?

flask_appbuilder/models/sqla/__init__.py
```diff
@@ -46,7 +45,7 @@ def get_tables_for_bind(self, bind=None):
         return result
 
 
-class ModelDeclarativeMeta(_BoundDeclarativeMeta):
+class ModelDeclarativeMeta(DefaultMeta):
```
dpgaspar: Can you explain this change?

dpgaspar:
Thank you @jmcarp , can you explain the changes on SQLA and ModelDeclarativeMeta? is it related to SQLAlchemy version bump?

flask_appbuilder/models/sqla/__init__.py
fordguo: > _BoundDeclarativeMeta is renamed to DefaultMeta; the old name will be removed in 3.0

see https://github.com/mitsuhiko/flask-sqlalchemy/blob/50944e77522d4aa005fc3c833b5a2042280686d3/CHANGES.rst#version-230
and
mitsuhiko/flask-sqlalchemy#546

flask_appbuilder/models/sqla/__init__.py
jmcarp: This is the current signature of `make_declarative` base as of flask-sqlachemy 2.3: https://github.com/mitsuhiko/flask-sqlalchemy/blob/master/flask_sqlalchemy/__init__.py#L734

jmcarp:
I think @fordguo and I covered the changes. Is there anything else we should document or explain @dpgaspar ?

dpgaspar:
LGTM, although I don't like it much :P

jmcarp:
@dpgaspar: I pinned all versions to a range like you suggested in https://github.com/dpgaspar/Flask-AppBuilder/pull/814#issuecomment-426546110. Do you think we should use more restricted ranges?

dpgaspar:
No it's good, good work @jmcarp

PR #814, opened 2018-09-28
update the requirements txt to allow higher versions

SaurabhBajaj:
Currently, many packages in the setup.py are hardcoded to a specific version. Unless there is a breaking issue with higher version, we should let users install any of the minor versions of the package. This would prevent issues with piptools pinning of python library versions

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
         'colorama==0.3.9',
         'click==6.7',
         'Flask>=0.10.0,<0.12.99',
-        'Flask-Babel>=0.11.1',  # known issues with 0.11.2
+        'Flask-Babel==0.11.1',  # known issues with 0.11.2
         'Flask-Login>=0.3,<0.5',
         'Flask-OpenID>=1.2.5',
         'Flask-SQLAlchemy>=2.1',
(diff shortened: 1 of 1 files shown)


SaurabhBajaj:
@dpgaspar @mistercrunch

mistercrunch:
I agree that we should use ranges and blacklisting known-incompatibility issues. Personally I think `==` should come with a comment documenting the known issue. Being caught in the past is a pretty bad thing.

Though now we need to understand why were those pinned in the first place.  @dpgaspar do you agree with the general approach (use ranges, and try to >=), any insight on the pinned libs?

jmcarp:
All these dependencies are upgraded in #813, as well as a few more that have been causing conflicts in airflow. WDYT @SaurabhBajaj @mistercrunch ?

SaurabhBajaj:
great timing @jmcarp . I am fine with moving your PR forward and closing this one

dpgaspar:
I personally don't like unpinned dependencies. I've had some bad experiences with it, when using AWS autoscaling, scaling up can spawn broken instances. A better approach could be to frame the versions like:
```
Flask>=0.10.0,<0.12.99
```

jmcarp:
I think this PR can be closed in favor of #813. @dpgaspar, I took your suggestion and set maximum dependency versions in that PR.

dpgaspar:
@jmcarp , yes they are overlapping. closing it

Issue #581, opened 2017-08-09
Python 3.4+ compatibility issue


myliyifei:
I found this issue was about Python 3.4+ compatibility which moves reduce function to "functools" module, can you fix that?

[https://stackoverflow.com/questions/8689184/nameerror-name-reduce-is-not-defined-in-python](url)

builtins.NameError
NameError: name 'reduce' is not defined

Traceback (most recent call last)
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\app.py", line 1997, in __call__
return self.wsgi_app(environ, start_response)
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\app.py", line 1985, in wsgi_app
response = self.handle_exception(e)
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\app.py", line 1540, in handle_exception
reraise(exc_type, exc_value, tb)
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\_compat.py", line 33, in reraise
raise value
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\app.py", line 1982, in wsgi_app
response = self.full_dispatch_request()
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\app.py", line 1614, in full_dispatch_request
rv = self.handle_user_exception(e)
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\app.py", line 1517, in handle_user_exception
reraise(exc_type, exc_value, tb)
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\_compat.py", line 33, in reraise
raise value
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\app.py", line 1612, in full_dispatch_request
rv = self.dispatch_request()
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask\app.py", line 1598, in dispatch_request
return self.view_functions[rule.endpoint](**req.view_args)
File "C:\Users\myliyifei\VENV\flask-appbuilder\lib\site-packages\flask_appbuilder\security\decorators.py", line 26, in wraps
return f(self, *args, **kwargs)
[...]

PR #823, opened 2018-10-05
Bugfix/reduce fixes issue #581

dolfandringa:
Python 3.4+ moved reduce function to functools


dolfandringa:
This fixes issue #581

PR #826, opened 2018-10-08
New Greek Translation

pliroforikos:
Greek Translation and compiled file.

I used German file and did the translation from it


dpgaspar:
ok, looks good, Thank you!

PR #827, opened 2018-10-12
Fix: Only one end of the one-to-one relation returned the right result

dolfandringa:
The SQLAInterface.is_relation_one_to_one only returned true for one end of a one-to-one relation, not for the other end. This is because sqlalchemy reports the relation with direction=ONETOMANY or direction=MANYTOONE (depending on the side of the relation). The previous code only checked the ONETOMANY side, not the MANYTOONE. This is now fixed (including a test).

### DIFF
--- a/flask_appbuilder/models/sqla/interface.py
+++ b/flask_appbuilder/models/sqla/interface.py
@@ -223,6 +223,8 @@ class SQLAInterface(BaseInterface):
     def is_relation(self, col_name):
         try:
             return isinstance(self.list_properties[col_name], sa.orm.properties.RelationshipProperty)
+        except KeyError:
+            return False
         except Exception as e:
             log.warning("An error occured when determining the relation")
             log.exception(e)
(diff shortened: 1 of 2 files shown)


dolfandringa:
Sorry, Fixing this now. Bad me, not running tests properly. Error is with getting the widget for the related_views that have a one-to-one relation.

dolfandringa:
So, now all tests work. THe problem was a lot harder to fix than I thought, but got it. The thing that made it hard were the blanket except statements in there without any mention of the exception.

Checking relationship directions is a lot more complicated in sqlalchemy than it seems initially. The code was working because one-to-one relations weren't being used. But the function testing for them was properly broken, and as soon as I started fixing that one, it was clear that the others were incorrect too.

To check relationships, you need to check both the direction (MANYTOONE/ONETOMANY/MANYTOMANY) and the `uselist` attribute of the RelationshipProperty. The problem is that for MANYTOONE you need to check the `uselist` attribute of the remote side, which may not always exist (no backref/back_populates). In reality, sqlalchemy should probably just make sure the direction.name attribute is correct. Now, one-to-one relations report as MANYTOONE or ONETOMANY depending on the side of the relation.

dpgaspar:
Nice work, let me check it, this should be included on 1.12.2

dpgaspar:
Hi @dolfandringa, please submit it to master also.

dolfandringa:
> Hi @dolfandringa, please submit it to master also.

Done

TeoGoddet:
Hello, it seems thos change are not in current releases, is there a reason ?

PR #828, opened 2018-10-16
Fix list and form widget rendering html InLine

mistercrunch:
It appears that since
https://github.com/dpgaspar/Flask-AppBuilder/pull/797 , there have been
issues rendering inlines that show as HTML.

This quick fix addresses it, and a quick test shows that <script> tags
get rendered as strings so things are still secure following this fix.

bildschirmfoto 2018-10-16 um 15 45 54

### DIFF
--- a/flask_appbuilder/templates/appbuilder/general/widgets/base_list.html
+++ b/flask_appbuilder/templates/appbuilder/general/widgets/base_list.html
@@ -26,7 +26,7 @@
     {% block end_content scoped %}
     {% endblock %}
 
-    {{ lib.action_form(actions,modelview_name) }}
+    {{ lib.action_form(actions, modelview_name) }}
 
     <script language="javascript">
         var modelActions = new AdminActions();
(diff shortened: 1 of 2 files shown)


mistercrunch:
@dpgaspar this is probably a top priority and it would nice to quickly release `0.18.1` as a fix.

dpgaspar:
1.12.1 is out!

lahorekid:
> 
> 
> 1.12.1 is out!

downloaded flask -> pip install flask -appbuilder 

but

 List Columns are still HTML on Superset :(

dolfandringa:
@mistercrunch can you confirm that @lahorekid is saying. Is there still an issue? If so, I'll dive into it, but don't want to start from scratch investigating it just to find out it's actually fixed. :smile:

dpgaspar:
Great @dolfandringa, can we be more specific? I can't reproduce this, have not tried it with superset

lahorekid:
> 
> 
> Great @dolfandringa,
> 
> @lahorekid, can you be more specific? I can't reproduce this, have not tried it with superset

upgraded to 1.12.1 by specifying the version...it has worked.

dolfandringa:
Ok, good to know @lahorekid. Thanks for following up on this.

PR #829, opened 2018-10-17
New, added support for additional filtering in LDAP search

svchiluvuri:
**Advanced filtering in LDAP search**

I added a new property called `AUTH_LDAP_SEARCH_FILTER` to specify additional filtering when searching for a LDAP user. 

Here's an example of how it can be used. 

```python
AUTH_LDAP_SEARCH_FILTER = "(&(objectCategory=Person)(memberOf=CN=Sales Team,OU=Sales Org,OU=Global Sales Org,OU=Global Corp,DC=corp,DC=com)"
```

If this property is set, it will be prepended to the default `filter_str` allowing to perform complex search filtering. I think this will be useful in two scenarios. 

1. Avoiding timeouts / connection errors when searching large directories by narrowing down search scope
2. Limiting access to users of a particular group within an organization (by checking if the user is a `memberOf` that group)

Sunil


dolfandringa:
Thanks. This looks like a good idea to me. But it could do with some tests. It may not be a trivial task, but this may help: https://mockldap.readthedocs.io/en/latest/ (it's been a long time since I did anything with ldap though).  A good test would be where you compare the LDAP search arriving at your mock ldap server and see if the search string is correct with and without the AUTH_LDAP_SEARCH_FILTER config attribute. Is that something you would be willing/able to add?

svchiluvuri:
I did delete the wrong line, fixed now. 

I will look into writing the tests. 

Thanks.

svchiluvuri:
Here's my (rather poor) attempt at writing tests for LDAP search and filtering.

dolfandringa:
This is looking good @svchiluvuri . Thanks a lot. Just one more request if it isn't too much to ask. It isn't very well documented with mockldap, but you can use 
```
self.assertEqual(self.ldapobj.methods_called(with_args=True), "<whatever you're expecting>")
```

to test whether the right ldap queries have been called. I didn't try it on your code yet, but do you think this could be added? That would allow you to test how the ldap queries change depending on your config parameter. If that could be added you have the ideal contribution (that I'd gladly show others as a good example).  If you don't have time, don't worry and I'll probably add it myself.

svchiluvuri:
Added check to confirm right LDAP queries are being called for both scenarios. Thanks @dolfandringa for being patient. This has be a good learning experience for me.

dolfandringa:
I'm happy to show you. Especially because you're persevering and helping.
Sorry for being annoyingthough, but what I actually meant is not to just test which methods were called, but also what arguments were sent to the search. It would allow you to test the actual search. If I am correct (not 100% sure) you would then see that with your option, the search arguments are different than without your search option. That is what the `with_args=True` part does. 

Here is an example (that should probably be documented better with mockldap)

```
import ldap
from mockldap import MockLdap

top = ('o=test', {'o': ['test']})
example = ('ou=example,o=test', {'ou': ['example']})
other = ('ou=other,o=test', {'ou': ['other']})
manager = ('cn=manager,ou=example,o=test', {'cn': ['manager'], 'userPassword': ['ldaptest']})
alice = ('cn=alice,ou=example,o=test', {'cn': ['alice'], 'userPassword': ['alicepw']})
bob = ('cn=bob,ou=other,o=test', {'cn': ['bob'], 'userPassword': ['bobpw']})

directory = dict([top, example, other, manager, alice, bob])
mockldap = MockLdap(directory)
mockldap.start()
ldapobj = mockldap['ldap://localhost/']

conn = ldap.initialize('ldap://localhost/')
conn.simple_bind_s('cn=alice,ou=example,o=test', 'alicepw')
results = conn.search_s('ou=example,o=test', ldap.SCOPE_ONELEVEL, '(cn=*)')
 
ldapobj.methods_called(with_args=True)

```

Output:
``` 
[('initialize', ('ldap://localhost/',), {}),
 ('simple_bind_s', ('cn=alice,ou=example,o=test', 'alicepw'), {}),
 ('search_s', ('ou=example,o=test', 1, '(cn=*)'), {})]
```

See the arguments in there? I would assume that your configuration option would change the search query right?

dpgaspar:
the mockldap test will be really helpful, great work. I probably need to change LDAP TLS configuration,    going to branch and submit a PR, you will probably need to rebase this one later.

svchiluvuri:
@dpgaspar That sounds good. 

@dolfandringa I think my code does exactly what you're describing. 

I defined the expected calls for both scenarios like below. 

[**Simple search expected calls**](https://github.com/dpgaspar/Flask-AppBuilder/blob/1545fcbdd37e36d25f7651ae49c68ef1f4719f7e/flask_appbuilder/tests/test_ldapsearch.py#L71-L73)

[**Search and filter expected calls**](https://github.com/dpgaspar/Flask-AppBuilder/blob/1545fcbdd37e36d25f7651ae49c68ef1f4719f7e/flask_appbuilder/tests/test_ldapsearch.py#L86-L88)

Then in the assertEqual I compare the method calls to these variables. 

[**Simple search assertEqual**](https://github.com/dpgaspar/Flask-AppBuilder/blob/1545fcbdd37e36d25f7651ae49c68ef1f4719f7e/flask_appbuilder/tests/test_ldapsearch.py#L76)

[**Search and filter assertEqual**](https://github.com/dpgaspar/Flask-AppBuilder/blob/1545fcbdd37e36d25f7651ae49c68ef1f4719f7e/flask_appbuilder/tests/test_ldapsearch.py#L91)

Can you take a second look and let me know if I'm missing something. 

Thanks.

Issue #831, opened 2018-10-19
Flask-AppBuilder installs an examples package alongside flask_appbuilder. This can overwrite local packages.


elvijs:
## Summary

`Flask-AppBuilder` installs an `examples` package alongside `flask_appbuilder`, which may overwrite local `examples` packages.

### Environment

Flask-Appbuilder version: 1.12.0

pip freeze output: 
```
Babel==2.6.0
click==6.7
colorama==0.3.9
defusedxml==0.5.0
examples==1.0
Flask==0.12.4
Flask-AppBuilder==1.12.0
Flask-Babel==0.11.1
Flask-Login==0.4.1
Flask-OpenID==1.2.5
Flask-SQLAlchemy==2.1
Flask-WTF==0.14.2
ItsDangerous==1.0.0
Jinja2==2.10
MarkupSafe==1.0
pkg-resources==0.0.0
python-dateutil==2.7.3
python3-openid==3.1.0
pytz==2018.5
six==1.11.0
SQLAlchemy==1.2.12
Werkzeug==0.14.1
WTForms==2.2.1
```

Project setup: a local project `myapp` with a local `examples` package, which in turn contains a `somestuff` module.

Set up the virtual environment in the following way:
1. `pip install -e examples` (this is my local project's examples package).
2. `pip install FlaskAppBuilder==1.12.0`.

### Describe the expected results

Since I have installed the local `examples` package into my venv, I expect to be able to run `import examples.somestuff` without any error.

```python
import examples.somestuff
```

### Describe the actual results

I actually get `ImportError: No module named 'examples.somestuff'`. This is because the Flask-AppBuilder has installed *its own* `examples` package, which has overwritten my local package.

### Steps to reproduce

Download and decompress the attached files: [flask-appbuilder-example-repro.tar.gz](https://github.com/dpgaspar/Flask-AppBuilder/files/2495096/flask-appbuilder-example-repro.tar.gz). Then follow the `README.md` instructions

dolfandringa:
Thanks for reporting this @elvijs. I see it too. It think this is caused by the fact that the "examples" directory of flask appbuilder is a module (has an __init__.py file). Setuptools sees it as a separate package and installs it. I think we should remove the __init__.py from the examples directory. I'll try to remember to try this out after the weekend and create a PR for it (unless someone else beats me to it, which would be nice).

dolfandringa:
This issue has been fixed in pull request #832

PR #832, opened 2018-10-19
Fix: Removed the examples __init__ file

elvijs:
This solves https://github.com/dpgaspar/Flask-AppBuilder/issues/831


dolfandringa:
Thanks elvijs! Looks good to me.

PR #837, opened 2018-10-24
Fix, is_authenticated is a function

c4milo:


dolfandringa:
I guess you figured this out already, but with the newer flask_login it is not. See the "changes" for the new release.

c4milo:
@dolfandringa, yes! sorry for the noise. Just in case someone lands here as well. I ran into this issue with Apache Airflow 1.10, it installs flask-login 0.2.x and also installs flask-appbuilder 1.12.1. So naturally, flask-appbuilder fails because `is_authenticated` is not a boolean.

Issue #838, opened 2018-10-26
AttributeError raised with another error when using MongoDB


dolfandringa:
In a simple mongodb sample (see [the gist](https://gist.github.com/dolfandringa/ed36b01cf97dd3f27d72848dd01b9cbd)), an error is logged about a missing relation. The cause of the error is due to many-to-many relationships don't really fitting into the NoSQL paradigm, but the error is logged in FAB, which raises its own error: `AttributeError: 'CourseView' object has no attribute 'name'`. This is second error is caused by the log statement on line 741 of baseviews.py that logs ` related_view.name`, which indeed doesn't exist.

PR #839, opened 2018-10-26
Fix issue 838: Views has no attribute name error raised in log statement

dolfandringa:
Fixes issue #838. Change the log statement to get the actual name of the view through getting the class `.__name__`  attribute instead of `.name`


dolfandringa:
@dpgaspar I assume I shouldn't review my own pull requests right?

PR #848, opened 2018-11-06
New,add oauth_remote_app method

mlboy:


dpgaspar:
Can you be more specific about the use case of this, and add a description also

PR #862, opened 2018-11-27
Fix, Console: Use appbuilder variable instead of literal.

### LINKED ISSUE #861: Console --appbuilder input not used in import_application function
I believe the highlighted line below needs to use the variable `appbuilder`, and not a string literal.

https://github.com/dpgaspar/Flask-AppBuilder/blob/2186b939e22db93fbd078f8a93b180776b3ad936/flask_appbuilder/console.py#L35

```
def import_application(app_package, appbuilder):
    sys.path.append(os.getcwd())
    try:
        _app = __import__(app_package)
    except Exception as e:
        click.echo(click.style('Was unable to import {0} Error: {1}'.format(app_package, e), fg='red'))
        exit(3)
    if hasattr(_app, 'appbuilder'):
        return getattr(_app, appbuilder)
    else:
        click.echo(click.style('There in no appbuilder var on your package, you can use appbuilder parameter to config', fg='red'))
        exit(3)
```

nick-schultz:
The problem is that when appbuilder is in a different module,  one needs to use `fabmanager --appbuilder` to specify the variable,  however the import_application code does not use the appbuilder variable.  Instead it uses the string literal "appbuilder".  I believe this is a typo.

fixes  #861

### DIFF
--- a/flask_appbuilder/console.py
+++ b/flask_appbuilder/console.py
@@ -32,7 +32,7 @@ def import_application(app_package, appbuilder):
     except Exception as e:
         click.echo(click.style('Was unable to import {0} Error: {1}'.format(app_package, e), fg='red'))
         exit(3)
-    if hasattr(_app, 'appbuilder'):
+    if hasattr(_app, appbuilder):
         return getattr(_app, appbuilder)
     else:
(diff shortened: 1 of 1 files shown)


dpgaspar:
Thank you

Issue #818, opened 2018-10-04
OAuth logic for finding user seems broken


victornoel:
I couldn't login with google oauth even though the user that was created in the database with the correct email.

The doc says:
```
This info will be checked with the internal user (user record on User Model), first by username next by email.
```

So I would expect that if one of those matches, then it's alright.

But in practice, only the username is always checked because the google oauth answer always contains a username (which seems to be generated by FAB since it has a format like `google-xxxxxxxx` where `xxxxxx` is the id of the google user).

The incriminating code is:
```py
        if 'username' in userinfo:
            user = self.find_user(username=userinfo['username'])
        elif 'email' in userinfo:
            user = self.find_user(email=userinfo['email'])
        else:
            log.error('User info does not have username or email {0}'.format(userinfo))
            return None
```

I think that if the first `find_user` fails, email should be checked.

victornoel:
For the record, I'm using:
Flask-AppBuilder         1.11.1
Flask-OAuthlib           0.9.5
google-auth              1.5.1    
google-auth-httplib2     0.0.3    
google-auth-oauthlib     0.2.0

dolfandringa:
This indeed looks like a bug to me if indeed the intended behaviour is that either the username or email should match. Can you shine light on this @dpgaspar? If that is the intended behaviour, a PR (including unittest) that fixes this would be very welcome.

holandes22:
Any idea when can this fix be merged? it is also affecting the new Airflow UI (based on this project)
https://issues.apache.org/jira/browse/AIRFLOW-4153

keerthanaturlapati:
Hi, any idea when will the fix be merged ?

apelliciari:
Hi, I have the same issue, please merge!

mgmorcos:
This issue is also affecting Apache Superset.

PR #863, opened 2018-11-27
Fix: Oauth security search is not sufficiently comprehensive

danrubenstein:
As described in #818, in the case where Google Oauth is used (and I think generally in Oauth all around), the current logic is to search for a username if the `username` key exists in the userinfo dictionary, and only if the `username` key does not exist is a search done with email. 

This fix searches for a user by username, and if the result of find_user from that search returns a false-y value, it then proceeds to search by email.

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -821,13 +821,17 @@ class BaseSecurityManager(AbstractSecurityManager):
             :userinfo: dict with user information the keys have the same name
             as User model columns.
         """
-        if 'username' in userinfo:
-            user = self.find_user(username=userinfo['username'])
-        elif 'email' in userinfo:
-            user = self.find_user(email=userinfo['email'])
-        else:
+        user = None
+
+        if 'username' not in userinfo and 'email' not in userinfo:
             log.error('User info does not have username or email {0}'.format(userinfo))
             return None
+
+        if 'username' in userinfo:
+            user = self.find_user(username=userinfo['username'])
+        if not user and 'email' in userinfo:
+            user = self.find_user(email=userinfo['email'])
+
         # User is disabled
         if user and not user.is_active:
             log.info(LOGMSG_WAR_SEC_LOGIN_FAILED.format(userinfo))


danrubenstein:
Hi @dpgaspar @dolfandringa,

I'm having trouble requesting a review, but would you be willing to look at this and see if it fixes #818?Also do feel free to let me know if there's anything else you'd like from me here. 

Thanks for your time and consideration.

mistercrunch:
LGTM

danrubenstein:
I am happy to close this.

holandes22:
@danrubenstein I'm curious how did you ended up solving this? I had to make a fork of FAB to use in our Airflow 1.10 deployment to make it work with google auth due to this

steve12d:
Hi all, I also ran into the same issue and the [fix](https://github.com/danrubenstein/Flask-AppBuilder/commit/04f84b61bc170e12b8de6e5abef63de8fb644674?diff=split) used by @danrubenstein worked for me.

danrubenstein:
@holandes22 - yeah, like you I think, I wanted to do RBAC with Airflow 1.10. The way we had solved with before trying to upgrade to Airflow 1.10 was with an Airflow Plugin that used a custom Flask BluePrint that would only interact with the creative / destructive paths.

This led to a very simple permission model that used the "owner" property of a DAG which would let the owner (as determined by name), plus some set of super users (as emails, although you could do this in another way) trigger, pause, etc a DAG, but everyone who could access the platform would have the equivalent of read-only access.

I might be able to dig up the source if you're interested - we're still using it instead of Flask-AppBuilder + RBAC, even after upgrading to 1.10.

holandes22:
Thank you for the offering @danrubenstein! but it's fine, we ended up using your [fix](https://github.com/danrubenstein/Flask-AppBuilder/commit/04f84b61bc170e12b8de6e5abef63de8fb644674?diff=split) in our own fork

It works great, but I rather this be merged into upstream instead of having to fork the repo

holandes22:
@dpgaspar any chance this can be re-opened and merged? We need to rely on a fork to keep using Airflow 1.10 with FAB

kdwyer:
@dpgaspar it looks as if this issue was automatically closed after you reopened it.  It would be good to have this change merged, if you find it satisfactory.

psheth-twist:
@dpgaspar We would love to upgrade to latest Airflow if this fix can be merged please :)

danrubenstein:
I've lost the thread here - if someone else wants to pick it up, go for it.

PR #882, opened 2018-12-29
Fix, Missing README.rst for example

sn1p3r46:
README.rst was missing for the specific example, No code affected.


dpgaspar:
Thank you!

PR #883, opened 2018-12-30
Support jinja2 StrictUndefined.

jmcarp:
So that apps using appbuilder can configure jinja to error on undefined variables, explicitly test whether template variable are defined rather than treating undefined as falsy.

### DIFF
--- a/flask_appbuilder/baseviews.py
+++ b/flask_appbuilder/baseviews.py
@@ -455,7 +455,6 @@ class BaseModelView(BaseView):
         return ret
 
 
-
 class BaseCRUDView(BaseModelView):
     """
         The base class for ModelView, all properties are inherited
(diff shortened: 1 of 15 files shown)


jmcarp:
@dpgaspar PTAL when you have time.

feng-tao:
PTAL @mistercrunch , Max, I recall you are also a maintainer of the repo?

jmcarp:
Ping @dpgaspar @mistercrunch

feng-tao:
hey Max @mistercrunch , when you have time, could you take a look?

jmcarp:
@mistercrunch: I only added `jinja2.StrictUndefined` in unit tests to verify that all template variables are defined, so users shouldn't be affected.

jmcarp:
Ping @mistercrunch @dpgaspar when you have time.

Issue #884, opened 2018-12-31
list view ordering problem of related model with dotted notation


deeagle001:
Hello, I would like to report a bug or undocumented restriction regarding the ordering of related field on show view.

### Environment

Flask-Appbuilder version:
1.12.2

pip freeze output:
Nothing extra. Only the dependencies of the Flask-AppBuilder itself.

### Describe the problem

The documentation mentioned that related field can be ordered with the dotted notation at the end of the linked chapter:
https://flask-appbuilder.readthedocs.io/en/latest/quickhowto.html?highlight=order#define-your-views-views-py

However, the note does not mention the restriction which I found in the Issue related to the feature:
https://github.com/dpgaspar/Flask-AppBuilder/issues/51

@ben-github mentions the real restriction, which is more than I thought for the first time.
He mentioned, that the relationship field shall match with the related table name.
In most of the cases, the user can overcome this by renaming the desired `relationship` field.
In my case, the model has two foreign keys (relationships) to the same related model.
This case is impossible to be described with the given restriction.

Additionally, if you try to rename at least one of the `relationship` fields to satisfy the restriction, the sqlalchemy raises an error, that it could not figure out the ON clause for the JOIN statement, as the model has more than one foreign key for the same related table.
This happens in the models/sqla/interface.py in the `query` function, where it appends the join statements on the query in case of dotted notation.

deeagle001:
bump

dpgaspar:
Merged to master, thank you @deeagle001

PR #885, opened 2018-12-31
Fix, list view ordering problem of related model with dotted notation

deeagle001:
Fix for https://github.com/dpgaspar/Flask-AppBuilder/issues/884 issue.
Please read it for problem details.

In my fix, I used the `relationship` contained join statement (`primaryjoin`) for dotted notation query joins.

This deprecates the restriction to use the same field name for `relationship` as the related table's name, and allows the user to use more than one relation on the same model, whilst they can be used in dotted notation and be ordered.

`isouter` is needed for not to filter out the results in case the `relationship` is optional (`nullable`).
Wihtout it, the row number indicates the found results, but because of the join statement (inner), the list will not show the ones, where the `relationship` is null.


dpgaspar:
Hi,

This looks good, can you rebase and update?

deeagle001:
Sure. I will need three-way merge for this. The update seems not trivial for the first sight.
Thank you in advance for your patience!

dpgaspar:
Nice! sorry for the delayed answer on this one

deeagle001:
@dpgaspar I made the update and it seems travis accepts it.
Please review and accept the changes again.

dpgaspar:
Ok, looks good

PR #911, opened 2019-02-15
Fix, update manager.py to fix missing missing session rollback() when user add fails.

mpvismer:
Added missing rollback() - no clear reason why this was left out.


dpgaspar:
True, thank you

dpgaspar:
Just released 1.12.4 with this

Issue #672, opened 2017-12-19
MySQL server has gone away


xing393939:
I use uwsgi to serve the app，then i got this error

I try this solutions:
1，set SQLALCHEMY_POOL_SIZE=0 or SQLALCHEMY_POOL_RECYCLE lower then wait_timeout of mysqld
2，close the db.session after flask close a connection
3，SQLALCHEMY retry connect if first connection failed

but do not work，any ideas ?

PR #923, opened 2019-03-14
[Fix] SQLAlchemy enabling the pre-ping feature

john-bodley:
This PR attempts to resolve the "MySQL server has gone away" which is caused by an idle timeout. This was raised in Fixes https://github.com/dpgaspar/Flask-AppBuilder/issues/672 and we've also noticed it in our Superset deployment which uses Flask-AppBuilder. 

The solution is based on enabling the SQLAlchemy "pre-ping" feature which was suggested [here](https://stackoverflow.com/a/45078959) and uses the pattern [here](https://github.com/mitsuhiko/flask-sqlalchemy/issues/589) for extending the Flask SQLAlchemy class. Note we've seen this issue with other Flask applications and this proposed fix seems to remedy the problem.

Fixes https://github.com/dpgaspar/Flask-AppBuilder/issues/672

to: @dpgaspar @mistercrunch

### DIFF
--- a/flask_appbuilder/models/sqla/__init__.py
+++ b/flask_appbuilder/models/sqla/__init__.py
@@ -44,6 +44,11 @@ class SQLA(SQLAlchemy):
                 result.append(tables[key])
         return result
 
+    def apply_pool_defaults(self, app, options):
+        """Apply the connection pool defaults enabling the 'pre-ping' feature."""
+        super(SQLA, self).apply_pool_defaults(app, options)
+        options['pool_pre_ping'] = True
+
 
 class ModelDeclarativeMeta(DefaultMeta):
     """
@@ -87,4 +92,3 @@ class Model(object):
     This is for retro compatibility
 """
 Base = Model
-


dpgaspar:
Hi @john-bodley,

Have you tried tweaking SQLALCHEMY_POOL_RECYCLE? take a look at: http://flask-sqlalchemy.pocoo.org/2.3/config/#timeouts
You can also (if possible) change the connect_timeout on mariadb/mysql if this is your backend.

john-bodley:
@dpgaspar per [this](https://stackoverflow.com/questions/6471549/avoiding-mysql-server-has-gone-away-on-infrequently-used-python-flask-server) thread it seems like the 2018+ approach is to use the pre-ping feature rather than `SQLALCHEMY_POOL_RECYCLE`?

Also per the SQLALchemy documenation regarding pre-pinging,

> The approach adds a small bit of overhead to the connection checkout process, however is otherwise the most simple and reliable approach to completely eliminating database errors due to stale pooled connections. 

I wonder if it makes more sense to try to get this into `Flask-SQAlchemy` instead.

john-bodley:
Actually it nows seems like Flask-SQLAlchemy is working on allowing one to configure the engine options per https://github.com/pallets/flask-sqlalchemy/pull/684. That seems like a better approach than this PR which was applying the logic from https://github.com/pallets/flask-sqlalchemy/issues/589#issuecomment-361075700.

dpgaspar:
I agree it makes much more sense on flask-sqlalchemy, going to close this then.

PR #929, opened 2019-03-21
[api] New, CRUD RESTful API

dpgaspar:
Hi,

This is a new feature that implements an automatic CRUD REST API. Using JWT for authentication, and Rison for URI argument parsing. One of the goals is to make it simple to migrate, or use side by side with your ModelViews classes. The API response contains enough metadata so that a JS component is able to render Forms and Lists dynamically.

Main features:
- client side field selection
- client side metadata selection
- Filters, just like on ModelViews
- Ordering
- Pagination
- Internationalization
- Authentication support for DB and LDAP (but this time you don't have to opt)

Take a looks at the docs: https://flask-appbuilder.readthedocs.io/en/fab2/rest_api.html

Note: This is still a work in progress


docs/rest_api.rst
```diff
@@ -0,0 +1,951 @@
...
+
+Next let's change our newly created model (HTTP PUT)::
+
+    $ curl -XPUT http://localhost:8080/api/v1/group/1 -d \
+     '{"name": "Friends Changed"}' \
+     -H "Content-Type: application/json" \
+     -H "Authorization: Bearer $TOKEN"
+    {
+      "result": {
+        "name": "Friends Changed"
```
mistercrunch: Quick question here: if the model has many fields and I pass only a subset of those fields in my post, only those get updated? I remember I fixed the old update endpoint in earlier FAB to have this behavior.

docs/rest_api.rst
```diff
@@ -0,0 +1,951 @@
...
+        personal_celphone = Column(String(20))
+        contact_group_id = Column(Integer, ForeignKey('contact_group.id'), nullable=False)
+        contact_group = relationship("ContactGroup")
+        gender_id = Column(Integer, ForeignKey('gender.id'), nullable=False)
+        gender = relationship("Gender")
+
+        def __repr__(self):
+            return self.name
+
+        def some_function(self):
```
mistercrunch: Logically this can only work for arg-free functions. Maybe the implementation should force `@property` decorated methods only (?)

docs/rest_api.rst
```diff
@@ -0,0 +1,951 @@
...
+
+The filter will act on all HTTP endpoints, protecting delete, create, update and display
+operations
+
+Simple example using doted notation, FAB will infer the necessary join operation::
+
+    class ContactModelRestApi(ModelRestApi):
+        resource_name = 'contact'
+        datamodel = SQLAInterface(Contact)
+        base_filters = [['contact_group.name', FilterStartsWith, 'F']]
```
mistercrunch: oh cool, so this supports other derivatives of `flask_appbuilder.models.sqla.filters` too right? Meaning we can reuse the same classes as we use against `ModelView`

mistercrunch:
I did a quick pass mostly focussing on reading `docs/rest_api.rst`. This looks solid!

On the GET side of things I was conflicted between "OMG THIS IS GENIUS!" and "damn we're re-implementing a subset of GraphQL". On the mutations side + auth + security (CRUD / perm / filters), this offers things that are way beyond the scope of GraphQL

This opens up a clear path and sets the foundation to a next generation FAB.

docs/rest_api.rst
dpgaspar: Yes, it can work like HTTP PATCH

```
curl http://localhost:8080/api/v1/contact/1 \         
-H "Content-Type: application/json" \
-H "Authorization: Bearer $TOKEN"
{
  "description_columns": {}, 
  "id": "1", 
  "include_columns": [
    "contact_group", 
    "gender", 
    "name", 
    "address", 
    "birthday", 
    "personal_phone", 
    "personal_celphone"
  ], 
  "label_columns": {
    "address": "Address", 
    "birthday": "Birthday", 
    "contact_group": "Contact Group", 
    "gender": "Gender", 
    "name": "Name", 
    "personal_celphone": "Personal Celphone", 
    "personal_phone": "Personal Phone"
  }, 
  "result": {
    "address": "", 
    "birthday": "2019-03-22", 
    "contact_group": 1, 
    "gender": 1, 
    "name": "contact1", 
    "personal_celphone": "11111", 
    "personal_phone": "99999"
  }
}

curl -XPUT http://localhost:8080/api/v1/contact/1 -d \
'{"name": "contact1 changed"}' \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $TOKEN"
{
  "result": {
    "address": "", 
    "birthday": "2019-03-22", 
    "contact_group": 1, 
    "gender": 1, 
    "name": "contact1 changed", 
    "personal_celphone": "11111", 
    "personal_phone": "99999"
  }
}

```

docs/rest_api.rst
dpgaspar: That's a good idea! I wonder if partials could make sense here, if not I'll force it

docs/rest_api.rst
dpgaspar: It's using the same ground, but without the MVC. Yes, I'm aiming for enabling an easy migration path.

dpgaspar:
Thank you @mistercrunch,

It's a build up from some ideas you gave me!

But, lot's of work still needs to be done. I'm hoping the community likes it, I would say this would be a fantastic addition for **Superset**, **Airflow** and hundreds of other apps out there.

Any ideas, critics, PR's are very much welcome. Hey everyone, help is needed and highly appreciated!

NOTE.txt
```diff
@@ -0,0 +1,11 @@
+# LOGIN_DB
```
craig-rueda: Did you mean to commit this?

flask_appbuilder/security/decorators.py
```diff
@@ -2,12 +2,53 @@
...
+
+    def wraps(self, *args, **kwargs):
+        permission_str = "{}{}".format(PERMISSION_PREFIX, f._permission_name)
+        if self.appbuilder.sm.is_item_public(
+                permission_str,
+                self.__class__.__name__
+        ):
+            return f(self, *args, **kwargs)
+        verify_jwt_in_request()
+        login_user(self.appbuilder.sm.get_user_by_id(int(get_jwt_identity())))
```
craig-rueda: Probably not the right place for this, but we're assuming here that users will be identified by a unique `int` (I'm assuming a PKEY on the user table). Would it be a big lift to allow for strings (UUIDs) as user ids? Again, this isn't really the best place for this, but in terms of "best practices" in security, using a one-up counter for security-related ids is usually frowned upon.

flask_appbuilder/models/filters.py
```diff
@@ -138,6 +151,20 @@ def _add_filter(self, filter_instance, value):
...
+    def rest_add_filters(self, data):
+        """
+            Adds list of dicts
+        
+        :param data: list of dicts
+        :return:
+        """
+        for _filter in data:
+            filter_class = map_args_filter.get(_filter['opr'], None)
+            print("OPR {}".format(filter_class))
```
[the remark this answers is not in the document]
dpgaspar: it needs cleaning, not even log.debug. I like print because in the end there should be none left.

flask_appbuilder/security/api.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+            Login endpoint for the API returns a JWT
+        :return:
+        """
+        if not request.is_json:
+            return self.response_400(message="Request payload is not JSON")
+        username = request.json.get('username', None)
+        password = request.json.get('password', None)
+        provider = request.json.get('provider', None)
+        if not username or not password or not provider:
```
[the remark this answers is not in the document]
dpgaspar: No they are all required

flask_appbuilder/security/api.py
```diff
@@ -0,0 +1,126 @@
...
+        # Identity can be any data that is json serializable
+        access_token = create_access_token(identity=user.id)
+        return self.response(200, access_token=access_token)
+
+
+class UserApi(ModelRestApi):
+    resource_name = 'user'
+
+    label_columns = {
+        'get_full_name': lazy_gettext('Full Name'),
```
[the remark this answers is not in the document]
dpgaspar: completely, been all day DRYing and making tests. These are good candidates also

flask_appbuilder/security/api.py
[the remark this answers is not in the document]
craig-rueda: What I meant was that the message of "Missing required parameter" could be made a bit more specific, depending on the field that was missing. (i.e. "Username" missing, or "password" missing, etc...)

docs/rest_api.rst
mistercrunch: Yes the interface of writing a method that receives a SQLAlchemy query object and returns a new one (presumably with filters applied based on user attributes) works great for us.

dpgaspar: exactly!

flask_appbuilder/security/api.py
```diff
@@ -0,0 +1,126 @@
+from flask import request
+from flask_jwt_extended import create_access_token
+from flask_babel import lazy_gettext
+from ..views import expose
+from ..api import BaseApi, ModelRestApi
+
+
+class SecurityApi(BaseApi):
+
+    route_base = '/api/v1/security'
```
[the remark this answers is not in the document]
dpgaspar: Changed

flask_appbuilder/security/api.py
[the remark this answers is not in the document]
dpgaspar: Changed

flask_appbuilder/security/decorators.py
dpgaspar: I see what you mean, and yes, this is probably not the best place. but let's talk about it

flask_appbuilder/security/decorators.py
```diff
@@ -2,12 +2,53 @@
...
+
+        By default the permission's name is the methods name.
+    """
+    if hasattr(f, '_permission_name'):
+        permission_str = f._permission_name
+    else:
+        permission_str = f.__name__
+
+    def wraps(self, *args, **kwargs):
+        permission_str = "{}{}".format(PERMISSION_PREFIX, f._permission_name)
```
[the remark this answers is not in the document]
dpgaspar: The time is coming, going to deprecate python2 soon

examples/base_api/app/api.py
```diff
@@ -0,0 +1,50 @@
...
+from flask_appbuilder.api import BaseApi, expose, rison, safe
+from flask_appbuilder.security.decorators import protect
+from . import appbuilder
+
+
+class MyFirstApi(BaseApi):
+
+    resource_name = 'myfirst'
+    version = 'v2'
+    route_base = '/newapi/v3/nice'
```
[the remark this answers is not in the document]
dpgaspar: None...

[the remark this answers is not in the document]
dpgaspar: Fixed

flask_appbuilder/api/__init__.py
```diff
@@ -0,0 +1,1117 @@
...
+        Use this decorator to set a new merging
+        response function to HTTP endpoints
+
+        candidate function must have the following signature
+        and be childs of BaseApi:
+        ```
+            def merge_some_function(self, response, rison_args):
+        ```
+
+    :param func: Name of the merge function where the key is allower
```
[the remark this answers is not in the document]
dpgaspar: Fixed

PR #962, opened 2019-04-23
New, [views] add edit button to show page

CameronNemo:
Fixes a minor UX issue where a user might:

* open an entry
* decide they want to edit or delete it
* hit the back button so they can do so, then have to find the entry again

Unfortunately this introduces a non-trivial issue by not redirecting to the list page after an entry is deleted. Additionally, it is only implemented for one of the show views, even though it is likely needed for multiple.


dpgaspar:
Seems like travis build did not trigger, can you resubmit or push a new commit?

CameronNemo:
Will a force push not trigger it?

PR #967, opened 2019-04-27
New: Add logging options for ldap

villebro:
Currently FAB does not support logging for LDAP, making it difficult to identify problems during configuration. This PR adds trace options to `ldap.initialize()` call for logging as described here: https://www.python-ldap.org/en/latest/reference/ldap.html#ldap.initialize


dpgaspar:
Thank you @villebro, this looks good. Will probably be out on the next release, next week

villebro:
@dpgaspar is this ok for merging?

PR #969, opened 2019-04-29
[WiP] [security] New, override class permission name and converge perms

dpgaspar:
Make it possible to override class permission name and implement a previous permission name so it's possible to converge from previous permissions to new permissions on all roles

### DIFF
--- a/flask_appbuilder/api/__init__.py
+++ b/flask_appbuilder/api/__init__.py
@@ -220,6 +220,15 @@ class BaseApi(object):
                 base_permissions = ['can_get']
 
     """
+    class_permission_name = None
+    """
+        Override class permission name default fallback to self.__class__.__name__
+    """
+    previous_class_permission_name = None
+    """
+        If set security cleanup will remove all permissions tuples
+        with this name
+    """
     allow_browser_login = False
     """
         Will allow flask-login cookie authorization on the API
@@ -320,6 +329,8 @@ class BaseApi(object):
         self.apispec_parameter_schemas = self.apispec_parameter_schemas or dict()
         self._apispec_parameter_schemas = self._apispec_parameter_schemas or dict()
         self._apispec_parameter_schemas.update(self.apispec_parameter_schemas)
+        self.class_permission_name = (self.class_permission_name or
+                                      self.__class__.__name__)
         if self.base_permissions is None:
             self.base_permissions = set()
             for attr_name in dir(self):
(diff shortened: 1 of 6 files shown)


mistercrunch:
omg

flask_appbuilder/api/__init__.py
```diff
@@ -220,13 +220,37 @@ class ExampleApi(BaseApi):
                 base_permissions = ['can_get']
 
     """
+    class_permission_name = None
+    """
+        Override class permission name default fallback to self.__class__.__name__
+    """
+    previous_class_permission_name = None
```
mistercrunch: This could be a list but I don't think that is a necessity. If `class_permission_name` was to change twice over a few version, someone who have to make sure to bridge the versions.

dpgaspar: Yes, we need to be aware also that migrations can only go so far. If we converge two classes to the same name, it's not possible to automatically converge them back

dpgaspar: Actually I take that back, it's possible, this needs more work

dpgaspar: Done, we can now revert/expand permissions back after having compressing them, so forget my previous comment on not being possible to "automatically converge them back"

flask_appbuilder/security/manager.py
```diff
@@ -1147,6 +1147,93 @@ def security_cleanup(self, baseviews, menus):
...
+    @staticmethod
+    def _get_new_old_permissions(
+            method_permission_name,
+            previous_permission_name,
+    ):
+        ret = dict()
+        for method_name, permission_name in method_permission_name.items():
+            old_permission_name = previous_permission_name.get(method_name)
+            if old_permission_name:
+                ret['can_' + permission_name] = 'can_' + old_permission_name
```
dpgaspar: Need to replace this with PERMISSION_PREFIX

flask_appbuilder/security/manager.py
```diff
@@ -1147,6 +1147,93 @@ def security_cleanup(self, baseviews, menus):
...
+                            new_perm_name
+                        )
+        return state_transition
+
+    def security_converge(self, baseviews, menus):
+        state_transition = self._create_state_transition(baseviews, menus)
+        if not state_transition:
+            log.info("No state transitions found")
+            return
+        print(f"State {state_transition}")
```
dpgaspar: Remove print

PR #971, opened 2019-04-30
New, [views] add edit button to show page

CameronNemo:
Fixes a minor UX issue where a user might:

* open an entry
* decide they want to edit it
* hit the back button so they can do so, then have to find the entry again

### DIFF
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -225,12 +225,6 @@ class UserModelView(ModelView):
             appbuilder=self.appbuilder,
         )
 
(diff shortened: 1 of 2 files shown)


CameronNemo:
Bump

CameronNemo:
Is there a way to implement such a widget without significant code duplication?

dpgaspar:
Major code dup here, yes: flask_appbuilder/templates/appbuilder/general/widgets/show.html

CameronNemo:
Would you be fine if I just added a boolean flag to the ModelView class to toggle this functionality?

PR #976, opened 2019-05-03
Dependency Bump - Previous 0.1.0 version pinned here had an invalid character in its name

tmswfrk:
and was not being built properly when built from source. Owner of module
created 0.1.1 version for me and I need this updated here for use upstream.

For Fixes:

For new features:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,7 +27,7 @@ markupsafe==1.1.1         # via jinja2
 marshmallow-enum==1.4.1
 marshmallow-sqlalchemy==0.16.2
 marshmallow==2.18.0
-prison==0.1.0
+prison==0.1.1
 pyjwt==1.7.1
 pyrsistent==0.14.11       # via jsonschema
 python-dateutil==2.8.0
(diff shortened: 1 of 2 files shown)


dpgaspar:
Strange that Pypi has 0.1.1 but on @betodealmeida 's : https://github.com/betodealmeida/python-rison/blob/master/prison/__version__.py
we still have 0.1.0

This probably should not be pinned on setup.py can you change it:
prison>=0.1.0<1.0.0

betodealmeida:
@dpgaspar I published 0.1.1 yesterday because someone had a problem installing the `tgz`. For some reason the file in PyPI was called `prison-0.1.0-.tar.gz`, and I wasn't able to upload a file with the correct name because it said it already existed.

I just pushed my local repo, so if you check https://github.com/betodealmeida/python-rison/blob/master/prison/__version__.py it's now on 0.1.1.

dpgaspar:
@betodealmeida, cool! thank you for your time. 

BTW prison is great! (this sounds strange!) 

:)

tmswfrk:
Made the change as requested! Thank you.

PR #978, opened 2019-05-03
Fix, #406 Support url_prefix to run fab with a url base root other t…

### LINKED ISSUE #406: Running with a url base root other than '/'
I am trying to run an instance of flask app builder that has a url prefix deeper than the root. 

For example, instead of:

mydomain.com/index.html
mydomain.com/users/list/
...

It would be:

mydomain.com/somedir/somesubdir/index.html
mydomain.com/somedir/somesubdir/users/list/
...

It looks like I need to specify a route_base = '/somedir/somesubdir/' in BaseView and static_url_path='/somedir/somesubdir/static/' appropriately. But setting these causes the template includes to not be accessible. 

Has anyone found exactly which components need to be modified for this two work?

Thanks

mmyjona:
…han '/'

For Fixes:

For new features:


dpgaspar:
Tests ok but fails flake8, also we need a FAB_URL_PREFIX to support app factory pattern

mmyjona:
This pr also fix url redirect after login. 
Now the login page will redirect to next parameter passed in via url instead of the index page.
The prefixed_redirect is confusing when mixed with self.appbuilder.get_url_for_index, because the latter one is already prefixed. So one should use redirect(self.appbuilder.get_url_for_index) in this case. And external url in next parameter should not be prefixed_redirect.

mmyjona:
I need help to test this. Im currently using responder to mount the app to some url endpoint.

mmyjona:
just comfirm this issue is already solved in master.

PR #983, opened 2019-05-05
[docs] Fix, lack of @expose decorator in rest api example

caoxin1988:
The first REST API doesn't work if @expose decorator is missed.
So add it back into this doc, and it won't make readers like me confused.

For Fixes:

For new features:


dpgaspar:
thks!

PR #986, opened 2019-05-06
New, Feature Builtin Roles

dpgaspar:
New feature, to enable builtin config defined roles using regex.

Main features:
-------------------
- No need for FAB to keep updating the 'Admin' role. This role can be defined has:
```
FAB_ROLES = {
    "Admin": [
        [".*", ".*"]
    ]
}
```
- Builtin roles are evaluated first, so prevents extra queries to the DB on ``has_access`` method
- Regex on views and permissions, simplifies management
  Example of an read only role:
```
FAB_ROLES = {
    ...
    "ReadOnly": [
        [".*", "can_list"],
        [".*", "can_show"],
        [".*", "menu_access"]
    ]
}
```
- Builtin roles are readonly, and can be viewed on the same role management CRUD
Screenshot from 2019-05-06 18-57-44

- Builtin role names are inserted on the DB on FAB bootstrap if FAB_UPDATE_PERMS=True. They only insert the name, if the name exists will update DB with: ``role.permissions = []`` (TODO)


flask_appbuilder/security/manager.py
```diff
@@ -290,6 +291,12 @@ def create_jwt_manager(self, app) -> JWTManager:
         jwt_manager.user_loader_callback_loader(self.load_user)
         return jwt_manager
 
+    def create_builtin_roles(self):
```
craig-rueda: 👍

PR #988, opened 2019-05-07
Update advanced.rst

noxqs:
on page : https://flask-appbuilder.readthedocs.io/en/latest/advanced.html
under custom fields, the first code block, last line:
return Markup('<b>' + custom + '</b>')
should be
return Markup('<b>' + self.custom + '</b>')


dpgaspar:
@noxqs just the advanced.rst, please

PR #1007, opened 2019-05-17
[api] New, support custom property fields

dpgaspar:
Support for custom property fields on ``ModelRestApi``

support SQLA Model fields like:
```
class SomeModel(Model):
    ....
    @property
    def some_property(self):
        return "some"

### DIFF
--- a/flask_appbuilder/api/convert.py
+++ b/flask_appbuilder/api/convert.py
@@ -167,6 +167,10 @@ class Model2SchemaConverter(BaseModel2SchemaConverter):
             field = EnumField(enum_class, dump_by=enum_dump_by, required=required)
             field.unique = datamodel.is_unique(column.data)
             return field
+        # is custom property method field?
+        if hasattr(getattr(_model, column.data), 'fget'):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: flask_appbuilder/api/convert.py [hunk 1]
```diff
@@ -167,6 +167,10 @@ def _column2field(self, datamodel, column, nested=True, enum_dump_by_name=False)
             field = EnumField(enum_class, dump_by=enum_dump_by, required=required)
             field.unique = datamodel.is_unique(column.data)
             return field
+        # is custom property method field?
+        if hasattr(getattr(_model, column.data), 'fget'):
+            return fields.Str(dump_only=True)
```
betodealmeida: @dpgaspar can we change this to `fields.Raw` instead?

I noticed that in Superset we're returning strings for properties of the database, including for booleans. This means that `allows_subquery` (for example) could potentially be returned as the string "false", which is truthy in Javascript.

I tested with `Raw` here, and I noticed it returns `false` instead of `"false"`, as expected, but I'm unsure if it would have other implications.

PR #1016, opened 2019-05-22
[api] patch version 1.x trunk with make api csrf exempt if csrf exists

jkornblum:
This is to duplicate changes made in #1013 into the v1.x trunk.

I copied changes from that PR and removed a few unit tests that were failing. I assume because the api is not the same from v1 to v2. These were the unit tests removed due to failures. All other tests were passing.

```
test_auth_builtin_roles
test_create_item_mm_field
test_permission_override
test_permission_converge_compress
test_permission_converge_expand
```


dpgaspar:
Tests are still failing,you can remove all test changes from the patch or fix them.

jkornblum:
Will TravisCI start automatically? I don't see it starting after pushing a few minutes ago.

dpgaspar:
sorry, but why are you adding pyyaml?

jkornblum:
Tests failed locally without adding `pyyaml` which I saw was a dep in 2.x trunk.
This file failed to import without it, `flask_appbuilder/base.py` when it tried to import line 5 `from .api.manager import OpenApiManager`

If you look at the `flask_appbuilder/api/__init__.py` line 15 it tries to import module yaml `import yaml`.
I don't know if you had another yaml lib installed? I ran `pip install -r requirements.txt` in a fresh python 3.6 environment and that line was failing. So I just added pyyaml from 2.x.

dpgaspar:
Yes, I understand, on 1.13.0 we have on setup.py:
'apispec[yaml]>=1.1.1<2

but only `apispec` on `requirements.txt`, let's change the requirements.txt to ``apispec[yaml]==1.1.1``

jkornblum:
done, thank you.

PR #1020, opened 2019-05-24
Fix, Pull back more user details from Github API

StewartW:
Defaulting to empty strings for emails results in the some applications
that use FAB to reject users with the same empty email address.

When trying to use Airflow (which uses FAB for RBAC) when you have more than one user register via Github OAuth it fails due to returning an empty string instead of an email. This pull request aims to bring the Github userinfo in line with some of the other providers. It adds in first name, last name and email. 

Since the email returned by /user is the public email, it might be None. Lines 501 - 503 are used to return the first private email 

Github doesn't make names mandatory - so if there's no name available it will use a placeholder value or just set the first name if that's all that's available 

Requires: 'read:user' and 'user:email' scopes to work reliably.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -488,6 +488,19 @@ key is just the configuration for flask-oauthlib::
                 'request_token_url':None,
                 'access_token_url':'https://accounts.google.com/o/oauth2/token',
                 'authorize_url':'https://accounts.google.com/o/oauth2/auth'}
+        },
+        # NB: When using Github Enterprise - the base_url will become http(s)://<your_enterprise_domain>/api/v3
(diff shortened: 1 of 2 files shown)


dpgaspar:
Hi @StewartW,

Thank you, it would be great if you could update the oauth example with the github configuration.

StewartW:
> Hi @StewartW,
> 
> Thank you, it would be great if you could update the oauth example with the github configuration.

Sure thing 👍

jykae:
@StewartW @dpgaspar could we open up this again? This would be important for Apache Airflow as well to get at least email address from Github user. Or make it configurable what details are passed.

PR #1033, opened 2019-06-12
Fix, missing import in views example

ncrocfer:
The last example in **BaseViews** documentation uses the `flash` function from flask, but the import was missing.


dpgaspar:
Right! thank you

PR #1041, opened 2019-06-21
[deps] Fix, new prison does not depend on requests

dpgaspar:
Remove `requests` from the requirements since new `prison` does not need it anymore.

https://github.com/apache/incubator-superset/pull/7643


villebro:
Thanks @dpgaspar . Now if we can get a 2.1.5 release with this cherry we can finalize removal of `requests` from Superset.

Issue #1039, opened 2019-06-21
Enabling SQLAlchemy Joined Loading 


john-bodley:
Whilst debugging a performance issue in Superset @etr2460 discovered that the `@has_access` and `@has_api_access` FAB decorator lazily loads the [User.roles](https://github.com/dpgaspar/Flask-AppBuilder/blob/786a90828d7add3842332526608843db92345ea8/flask_appbuilder/security/sqla/models.py#L106) and [Roles.permission](https://github.com/dpgaspar/Flask-AppBuilder/blob/786a90828d7add3842332526608843db92345ea8/flask_appbuilder/security/sqla/models.py#L76) relationships resulting in a separate SQL query whilst iterating over the roles and role permission [here](https://github.com/dpgaspar/Flask-AppBuilder/blob/009f2c8cebb991):

```
2019-06-20 17:43:03,515 INFO sqlalchemy.engine.base.Engine SELECT 1
2019-06-20 17:43:03,515 INFO sqlalchemy.engine.base.Engine ()
2019-06-20 17:43:03,515 INFO sqlalchemy.engine.base.Engine BEGIN (implicit)
2019-06-20 17:43:03,517 INFO sqlalchemy.engine.base.Engine SELECT ab_user.id AS ab_user_id, ab_user.first_name AS ab_user_first_name, ab_user.last_name AS ab_user_last_name, ab_user.username AS ab_user_username, ab_user.password AS ab_user_password, ab_user.active AS ab_user_active, ab_user.email AS ab_user_email, ab_user.last_login AS ab_user_last_login, ab_user.login_count AS ab_user_login_count, ab_user.fail_login_count AS ab_user_fail_login_count, ab_user.created_on AS ab_user_created_on, ab_user.changed_on AS ab_user_changed_on, ab_user.created_by_fk AS ab_user_created_by_fk, ab_user.changed_by_fk AS ab_user_changed_by_fk 
FROM ab_user 
WHERE ab_user.id = ?
2019-06-20 17:43:03,517 INFO sqlalchemy.engine.base.Engine (1,)
2019-06-20 17:43:03,520 INFO sqlalchemy.engine.base.Engine SELECT ab_role.id AS ab_role_id, ab_role.name AS ab_role_name 
FROM ab_role, ab_user_role 
WHERE ? = ab_user_role.user_id AND ab_role.id = ab_user_role.role_id
[...]

PR #1044, opened 2019-06-23
perf: improve perf while checking perms

mistercrunch:
Found this easy minor perf improvement while digging around finding
solutions for #1039

Related thoughts:
* my intuition is that the best/easiest optmization would be to simply
  run a single query joining user, roles and perms checking for that
  single
  permission very quickly. The only downside of that is if `user.roles` is
  getting iterated over later in the request cycle, you have to pay the
  cost of loading it anyways. Maybe `_has_view_access` could have
  different mode of operations where it tries to fetch and cache things
  as opposed to the single check against the db.
* noticed `re.match` being used potentially a lot for static roles, we
  may want to re.compile and cache. This may be a premature
  optimization though, unclear...

related to: https://github.com/dpgaspar/Flask-AppBuilder/issues/1039

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -1058,6 +1058,9 @@ class BaseSecurityManager(AbstractSecurityManager):
 
     def _has_view_access(self, user, permission_name, view_name):
         roles = user.roles
+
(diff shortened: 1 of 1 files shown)


dpgaspar:
Good catch

Issue #1035, opened 2019-06-14
Swagger Docs not using Bearer Header


deckar01:
### Environment

Flask-Appbuilder version: 2.1.4

pip freeze output: N/A

### Describe the expected results

When an access token is added to the authorization module of the swagger docs, it should include it in the API requests as a bearer header.

### Describe the actual results

No bearer header was included in the API requests. Including the header in the curl command and running it manually works correctly, but this defeats the purpose of the interactive swagger page.

### Steps to reproduce

1. Setup a new project 
2. Add a model rest API
3. Enable swagger
4. Open the swagger page
5. Obtain an access token via the login API
6. Add the access token to the "authorize" section at the top of the swagger page
7. Make an API request that require authorization

deckar01:
Based on the info I found in https://github.com/swagger-api/swagger-ui/issues/3860, there are supposed to be security values in the _openapi response for actions that require authorization, but there are none. `components.securitySchemes` has `jwt` set, but none of the paths are configured with a `security` property that causes swagger to actually include the token in the request.

dpgaspar:
Hi @deckar01,

Thank you for reporting. Yes `_openapi` has:
"securitySchemes":{"jwt":{"bearerFormat":"JWT","scheme":"bearer","type":"http"}}

deckar01:
I think [`BaseApi.operation_helper()`](https://github.com/dpgaspar/Flask-AppBuilder/blob/d5fa14a579a8d9ce538ed02f592db4789d71ef9d/flask_appbuilder/api/__init__.py#L411) needs to introspect the operation method's `_permission_name` property, then set `security` on the operation's spec if needed.

dpgaspar:
Hi @deckar01,

PR #1046 will include the security object by path, this will be included on the next release

dpgaspar:
Released on 2.1.6

PR #1046, opened 2019-06-28
[openapi] Fix, include security on paths

dpgaspar:
Following issue #1035

Includes OpenAPI security object on each path if path needs authentication


deckar01:
I pip installed from this branch to test it locally. The jwt is correctly being supplied as a request header on the generated swagger docs page now. Also, operations that should be public do not send the auth header. 💯

PR #1048, opened 2019-06-28
[openapi] Fix, require refresh security token

### LINKED ISSUE #1047: JWT Refresh Specs Missing Args
### Environment

Flask-Appbuilder version: 2.1.5

pip freeze output: N/A

### Describe the expected results

The security API for refreshing a JWT token should advertise the ways it can accept a refresh token.

### Describe the actual results

There is currently no way for the generated swagger page to submit a valid refresh request, because it does not define any arguments in its specs.

Screen Shot 2019-06-28 at 9 57 21 AM
Screen Shot 2019-06-28 at 9 57 57 AM

### Steps to reproduce

1. Open the swagger API page
2. Submit a valid request to the security login endpoint
3. Add the access token in the swagger jwt auth form
4. Submit a request to the security refresh endpoint

### Notes

There don't seem to be any docs on how `@jwt_refresh_token_required` reads the refresh token. [flask-jwt-extended `_decode_jwt_from_request()`](https://github.com/vimalloc/flask-jwt-extended/blob/67de379bc2184fa042aa4202034f3572fb750293/flask_jwt_extended/view_decorators.py#L250-L257) is the underlying implementation. Advertising the request arg would get swagger working, but it would be good to declare as many of the accepted formats in the spec as possible.

After looking at the implementation of `_decode_jwt_from_request()`, the first place it looks for a token is the cookie. Automatically storing the tokens in the cookie would make authorizing requests (and refresh requests) much more convenient for API clients.

deckar01:
Fixes #1047 

Add a second JWT security schema for the refresh token. This adds a second input for the refresh token to the auth form in the interactive swagger docs.

Screen Shot 2019-06-28 at 11 46 04 AM

Set the security schema on the refresh operation to the refresh token schema. This adds a lock icon to the refresh operation and automatically includes the refresh token in the header of refresh requests.

Screen Shot 2019-06-28 at 11 58 00 AM

---

This allows the refresh API to be tested through the swagger docs and documents how the refresh API should be used.

Screen Shot 2019-06-28 at 11 47 56 AM

Screen Shot 2019-06-28 at 11 47 48 AM

Screen Shot 2019-06-28 at 11 48 35 AM

### DIFF
--- a/flask_appbuilder/security/api.py
+++ b/flask_appbuilder/security/api.py
@@ -30,6 +30,7 @@ class SecurityApi(BaseApi):
         super(SecurityApi, self).add_apispec_components(api_spec)
         jwt_scheme = {"type": "http", "scheme": "bearer", "bearerFormat": "JWT"}
         api_spec.components.security_scheme("jwt", jwt_scheme)
+        api_spec.components.security_scheme("jwt_refresh", jwt_scheme)
 
     @expose("/login", methods=["POST"])
     @safe
@@ -128,6 +129,8 @@ class SecurityApi(BaseApi):
               $ref: '#/components/responses/401'
             500:
               $ref: '#/components/responses/500'
+          security:
+            - jwt_refresh: []
         """
         resp = {
             API_SECURITY_REFRESH_TOKEN_KEY: create_access_token(


dpgaspar:
LGTM

PR #1073, opened 2019-07-31
[WiP] [api] HTTP method GET item can optionally show results for an edit form

dpgaspar:
If we want to dynamically generate an edit form (client side, using React for example) current API will only return results associated with `show_columns`


dpgaspar:
.

PR #1092, opened 2019-08-23
fix sql query containing exists in column clause

ThomasP0815:
This patch fixes the problem using flask-appbuilder in combination with a MSSQL database.


dpgaspar:
Great thks, I'll make some tests with PG and MySQL

PR #1095, opened 2019-08-27
Fix, OAuth - set fallback when next url in state is empty

paulvic:
If the `next` value in state is an empty string, then the `next_url` is also set to an empty string which breaks redirection (currently reproducible in the latest version of Superset).

See https://github.com/apache/incubator-superset/issues/7739

To reproduce:
 - Install latest version of Superset (https://github.com/apache/incubator-superset) and configure OAuth with Google
 - Login with Google Authentication

Observed:
 - Error appears:
...
```
File "/src/superset/superset/venv/lib/python3.6/site-packages/flask_appbuilder/security/views.py", line 677, in oauth_authorized
resp = self.appbuilder.sm.oauth_remotes[provider].authorized_response()
KeyError: 'login'
```

Expected:
 - Redirected to `/superset/welcome`

Notes:
With this fix in place, I've verified that user is redirected to `/superset/welcome` when there's no next value, and that they are also correctly redirected to the next url if it is set.


paulvic:
@dpgaspar - is the codecov/patch failure something I need to look into?

dpgaspar:
Hi @paulvic,

Thank you. 
It would be great if you could add a test for it

paulvic:
@dpgaspar - any idea when the next release will be? I'd like to pick this up in Superset ASAP.

Many thanks!

PR #1116, opened 2019-09-12
New: expose menu data as serializable objects

nytai:
exposes a method on the menu class to output the menu as list of dictionaries that can be converted to json.

### DIFF
--- a/flask_appbuilder/menu.py
+++ b/flask_appbuilder/menu.py
@@ -1,4 +1,8 @@
-from flask import url_for
+from flask import current_app, url_for
+
+from .api import BaseApi, expose
+from .basemanager import BaseManager
+from .security.decorators import permission_name, protect
 
 
 class MenuItem(object):
@@ -53,6 +57,31 @@ class Menu(object):
 
     def get_list(self):
         return self.menu
+
+    def get_data(self, menu=None):
+        menu = menu or self.menu
+        ret_list = []
+
+        for i, item in enumerate(menu):
+            if item.name == '-' and not i == len(menu) - 1:
+                ret_list.append('-')
+            elif not current_app.appbuilder.sm.has_access("menu_access", item.name):
+                continue
+            elif item.childs:
+                ret_list.append({
+                    "name": item.name,
+                    "icon": item.icon,
+                    "label": str(item.label),
+                    "childs": self.get_data(menu=item.childs)
+                })
+            else:
+                ret_list.append({
+                    "name": item.name,
+                    "icon": item.icon,
+                    "label": str(item.label),
+                    "url": item.get_url()
+                })
+        return ret_list
 
     def find(self, name, menu=None):
         """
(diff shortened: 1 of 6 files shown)


flask_appbuilder/menu.py
```diff
@@ -41,11 +41,12 @@ def __repr__(self):
 class Menu(object):
     menu = None
 
-    def __init__(self, reverse=True, extra_classes=""):
+    def __init__(self, reverse=True, extra_classes="", security_manager=None):
         self.menu = []
         if reverse:
             extra_classes = extra_classes + "navbar-inverse"
         self.extra_classes = extra_classes
+        self.security_manager = security_manager
```
dpgaspar: not needed, if we are calling get_data on a request context we can access the security manager with `current_app.appbuilder.sm`

flask_appbuilder/tests/test_menu.py
```diff
@@ -0,0 +1,77 @@
...
+        self.app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
+        logging.basicConfig(level=logging.ERROR)
+
+        self.db = SQLA(self.app)
+        self.appbuilder = AppBuilder(self.app, self.db.session)
+
+        sess = PSSession()
+
+        class PSView(ModelView):
+            datamodel = GenericInterface(PSModel, sess)
```
dpgaspar: This can be removed

flask_appbuilder/tests/test_menu.py
```diff
@@ -0,0 +1,77 @@
...
+    def test_menu_api(self):
+        """
+            REST Api: Test menu data
+        """
+        uri = '/api/v1/menu/data'
+        client = self.app.test_client()
+        rv = client.get(uri)
+        eq_(rv.status_code, 200)
+        data = rv.data.decode('utf-8')
+        eq_(data, '{"data":[]}\n')
```
dpgaspar: let's use `assertEqual` I'm changing everything on another PR

flask_appbuilder/menu.py
```diff
@@ -123,3 +151,54 @@ def add_separator(self, category=""):
             raise Exception(
                 "Menu separator does not have correct category {}".format(category)
             )
+
+
+class MenuApi(BaseApi):
+    resource_name = "menu"
+
+    @expose('/data', methods=["GET"])
```
dpgaspar: I personally prefer just GET `/api/v1/menu`

flask_appbuilder/menu.py
```diff
@@ -123,3 +151,54 @@ def add_separator(self, category=""):
...
+                                  type: string
+                                childs:
+                                  $ref: \
+        '#/paths/menu/data/get/reponses/200/content/application/json/schema/properties/data'
+                              required:
+                                - name
+                                - label
+                                - icon
+        """
+        return self.response(200, data=current_app.appbuilder.menu.get_data())
```
dpgaspar: Currently FAB uses `result` key instead of `data`

flask_appbuilder/menu.py
```diff
@@ -123,3 +151,54 @@ def add_separator(self, category=""):
...
+                                  type: string
+                                icon:
+                                  type: string
+                                url:
+                                  type: string
+                                childs:
+                                  $ref: \
+        '#/paths/menu/data/get/reponses/200/content/application/json/schema/properties/result'
+                              required:
+                                - name
```
dpgaspar: `name` and `url` are required. `label`, `icon` are optional

flask_appbuilder/menu.py
```diff
@@ -123,3 +151,54 @@ def add_separator(self, category=""):
             raise Exception(
                 "Menu separator does not have correct category {}".format(category)
             )
+
+
+class MenuApi(BaseApi):
+    resource_name = "menu"
+
+    @expose('/', methods=["GET"])
```
dpgaspar: public endpoint, is this intentional? it really does not return any data when no user is authenticated

flask_appbuilder/menu.py
```diff
@@ -123,3 +151,54 @@ def add_separator(self, category=""):
...
+                                - name
+                                - label
+                                - icon
+        """
+        return self.response(200, result=current_app.appbuilder.menu.get_data())
+
+
+class MenuApiManager(BaseManager):
+    def register_views(self):
+        self.appbuilder.add_api(MenuApi)
```
dpgaspar: Let's create a new option key for this, take a look at: https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/manager.py#L581

flask_appbuilder/menu.py
nytai: Correct, it returns an empty array, currently. However if a user decides to expose a menu for the public role we'd still want to serve that [limited] menu, right?

flask_appbuilder/menu.py
nytai: You mean expose a config to enable/disable this endpoint? something like `FAB_EXPOSE_MENU_API`?

flask_appbuilder/menu.py
dpgaspar: Right, yet we do have a public role for this. It's actually a decision between exposing an unauthenticated EP that anonymously return an empty list, or exposing an authenticated EP that returns 401

flask_appbuilder/menu.py
dpgaspar: yep

PR #1120, opened 2019-09-15
New, preparation for case-insensitive search

szetaa:

For Fixes:

For new features:

### DIFF
--- a/flask_appbuilder/models/filters.py
+++ b/flask_appbuilder/models/filters.py
@@ -20,6 +20,7 @@ class BaseFilter(object):
     model = None
     name = ""
     is_related_view = False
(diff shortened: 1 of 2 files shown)


szetaa:
Hi, when running FAB against Postgres, search is implicitly always case sensitive.
This proposal prepares the model to allow case insensitive search by using "ilike" instead of "like". 

As the default value is "True", the behavior will not change with this pull request.

If this feature is considered useful, I will make a proposal how to activate the case-insensitive search. 

Idea 1: Checkbox in "general/widgets/search.html"
Idea 2: Configuration

CameronNemo:
I would personally find a config key per-view to be most helpful.

PR #1124, opened 2019-09-18
[MVC CRUD] New, hide tabs when only 'Detail' (no related_views)

mistercrunch:
Also injecting new CSS classes so that we can do some CSS adjustments on the
CRUD forms.


dpgaspar:
Tested and looks good

Issue #1133, opened 2019-09-20
Marshmallow version is restrictive and out of date


mattdornfeld:
The Marshamallow version is pretty restrictive and is not up to date with the latest major release of Marshmallow. Any chance this can be updated to allow for 3.x.x releases of Marshmallow. This package is a dependency for Airflow, which is pretty widely used. The version restriction basically precludes using any other packages that use a new version of Marshmallow along with Airflow.

dpgaspar:
Left a comment on the PR

ecederstrand:
Left a comment in the PR - the version restriction in Airflow is gone.

dpgaspar:
Thank you, great news!

PR #1111, opened 2019-09-06
Fix: marshmallow-sqlalchemy update breaks py2 and 3.5 compatibility

mgasner:
marshmallow-sqlalchemy 0.19.0 is incompatible with py2 and py3.5 (syntax errors); Flask-AppBuilder's dependency on marshmallow-sqlalchemy breaks downstream projects on py2 and py3.5, including Apache Airflow.

This change repins the dependency in setup.py to `marshmallow-sqlalchemy>=0.16.1, <0.19.0` instead of `marshmallow-sqlalchemy>=0.16.1, <1`. (cf. the requirements.txt pin to `marshmallow-sqlalchemy==0.16.2`.

PR #1134, opened 2019-09-20
Fix: Updated Marshmallow dependency version

mattdornfeld:
The Marshmallow dependency version is restrictive and out of date. Since this package is included with Airflow, this basically precludes running any packages that depend on a newer version of Marshmallow along with Airflow. This update should fix this issue. I also filed an issue in the tracker system.

https://github.com/dpgaspar/Flask-AppBuilder/issues/1133

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
         "Flask-WTF>=0.14.2, <1",
         "Flask-JWT-Extended>=3.18, <4",
         "jsonschema>=3.0.1, <4",
-        "marshmallow>=2.18.0, <2.20",
+        "marshmallow>=2.18.0, <4.0.0",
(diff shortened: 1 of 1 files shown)


CameronNemo:
how does this relate to #1111 ?

dpgaspar:
marshmallow 3.X.X seems to be ok on FAB. So this looks good. but take a look at:
https://github.com/apache/airflow/blob/master/setup.py#L357, Airflow is still using FAB < 2. So this PR will not solve your issue

ecederstrand:
Airflow now has requirement on FAB ~2.2. Can we revisit this issue?

NilsJPWerner:
Any movement on this? It seems like everything is aligned to solve this issue now, as long as the marshmallow version bump is included in a 2.2  point release.

PR #1135, opened 2019-09-20
Fix autofocus login to make UX more seamless.

MarcusSorealheis:
Fix is a stretch. It's mostly an enhancement, but since the user experience of a security step can be improved, I consider it a fix. 

To reproduce the problem, and eliminate other variables, open the login page on an incognito or private browser. You will see that a user needs to more the cursor to the username. If they are there, they want to login most likely, so we should simplify the process for the user and focus on the username field in the login form.


dpgaspar:
Following #1132 going to close this one. Thank you nonetheless

PR #1153, opened 2019-10-03
New, update jquery to 3.4.1

### LINKED ISSUE #380: Update jQuery for security fix
We're using Caravel, which in turn depends on Flask-AppBuilder.

Unfortunately, a third party security audit flagged Flask-AppBuilder's bundled jquery version as unsafe due to known vulnerabilities.

I've reviewed the jquery 1.x release notes and I believe the audit was referring to the minifier bug in uglify-js detailed [here](https://diracdeltas.github.io/blog/backdooring-js/).  This was fixed in jquery 1.12 when they [updated the version of grunt-contrib-uglify](https://github.com/jquery/jquery/commit/835e9218beef8f0c559da314fac01cb85dacb740).

Anyway, would you be open to either updating the bundled jquery version to the latest 1.x release or accepting a PR that makes that update?

CameronNemo:
Closes #380


dpgaspar:
Hi @CameronNemo,

This is a most wanted change, but it seems to "break" the following:

Before:

image

Using JQuery 3.4.1:

image

dpgaspar:
Hi @CameronNemo, do you have the time to review this?

CameronNemo:
For some reason the `<tbody>` disappeared... I added it into the search widget template so it should be good now, I imagine. Will have to do some testing.

ryw:
Looks like jquery 3.4.1 is already in https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/static/appbuilder/js/jquery-latest.js

dpgaspar:
Thks @ryw

PR #1157, opened 2019-10-09
Fix, make single db call on menu get data

dpgaspar:
Improve performance on menu endpoint, make a single db call


dpgaspar:
@nytai making a single db call ^^^

Issue #1110, opened 2019-09-06
Oauth2 can not be logout in server


cdmikechen:
If you'd like to report a bug in Flask-Appbuilder, fill out the template below. Provide
any extra information that may be useful

### Environment

Flask-Appbuilder version:
Flask-AppBuilder 1.13.0, using in superset==0.28.1
pip freeze output:
pip3 list |grep Flask-AppBuilder
Flask-AppBuilder       1.13.0   

I used cas-oauth2(5.3) to test.
```
from flask_appbuilder.security.manager import AUTH_OAUTH
AUTH_TYPE = AUTH_OAUTH
OAUTH_PROVIDERS = [
    {   'name':'casOauth2',
        'token_key':'access_token', 
        'icon':'fa-address-card',   
        'remote_app': {
            'consumer_key':'xxx',  
            'consumer_secret':'xxx', 
            'request_token_params':{},
            'access_token_method':'POST',    
            'access_token_params':{     
                'client_id':'xxx',
                'client_secret':'xxx'
            },
            'access_token_headers':{    
                'Authorization': 'Basic Base64EncodedClientIdAndSecret'
            },
            'base_url':'http://localhost:8443/cas/oauth2.0/',
            'access_token_url':'http://localhost:8443/cas/oauth2.0/accessToken',
            'authorize_url':'http://localhost:8443/cas/oauth2.0/authorize'
        }
    }
]
```

### Describe the expected results
Tell us what should happen:
When finishing configuration in Flask-AppBuilder with oauth2, I login with user A, and then logout. After that I change user B to login, it should work.

### Describe the actual results
Tell us what happens instead:
When I logout, Flask-AppBuilder can not logout in oauth2 server. So that when I want to change another user B login, It login directly with the user A.

### Steps to reproduce
Maybe we need to support a method to logout in oauth2 server.

jeverling:
Hi @cdmikechen maybe the problem is just that Flask Appbuilder redirects to the homepage after logout, so you get logged in straight back with User A?

[flask_appbuilder/security/views.py#L477](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/views.py#L477)
```
class AuthView(BaseView):
    route_base = ""
    login_template = ""
    invalid_login_message = lazy_gettext("Invalid login. Please try again.")
    title = lazy_gettext("Sign In")

    @expose("/login/", methods=["GET", "POST"])
    def login(self):
        pass

[...]

PR #1160, opened 2019-10-13
New, Fixes #1154 by adding LOGOUT_REDIRECT_URL

### LINKED ISSUE #1154: Make redirect after logout configurable
Hi all,

I think it would be really useful if the URL that Flask-Appbuilder redirects to after logout would be configurable - akin to Django's `LOGOUT_REDIRECT_URL` setting.

Right now, the user is redirected to the hompage:

[flask_appbuilder/security/views.py#L477](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/views.py#L477)
```
class AuthView(BaseView):
    route_base = ""
    login_template = ""
    invalid_login_message = lazy_gettext("Invalid login. Please try again.")
    title = lazy_gettext("Sign In")

    @expose("/login/", methods=["GET", "POST"])
    def login(self):
        pass

    @expose("/logout/")
    def logout(self):
        logout_user()
        return redirect(self.appbuilder.get_url_for_index)
```

This might not be what is desired though, maybe the user should see a "thank you and see you soon page" or something like that, and it can also lead to problems with setups that use oauth2 for authentication, because the user will be logged straight back in if the oauth2 provider is configured to allow login without a prompt. I think this is what happens in #1110
So I think having an option to configure where users are redirected to after logout would be really useful, e.g.

```
    return redirect(
        self.appbuilder.app.config.get(
            "LOGOUT_REDIRECT_URL", self.appbuilder.get_url_for_index
        )
    )
```

jeverling:
This pull request addresses #1154, by redirecting to `LOGOUT_REDIRECT_URL` in the *logout* view if `LOGOUT_REDIRECT_URL` is set in the config, falling back to `self.appbuilder.get_url_for_index`.
It also adds `LOGOUT_REDIRECT_URL` to the **Configuration keys** table in `docs/config.rst`, and adds a unit test.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -194,6 +194,8 @@ Use config.py to configure the following parameters. By default it will use SQLL
 |                                   | the existing languages with the countries  |           |
 |                                   | name and flag                              |           |
 +-----------------------------------+--------------------------------------------+-----------+
+| LOGOUT_REDIRECT_URL               | The location to redirect to after logout   |   No      |
++-----------------------------------+--------------------------------------------+-----------+
 | FAB_API_SHOW_STACKTRACE           | Sends api stack trace on uncaught          |   No      |
 |                                   | exceptions. (Boolean)                      |           |
 +-----------------------------------+--------------------------------------------+-----------+
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -477,7 +477,11 @@ class AuthView(BaseView):
     @expose("/logout/")
     def logout(self):
         logout_user()
-        return redirect(self.appbuilder.get_url_for_index)
+        return redirect(
+            self.appbuilder.app.config.get(
+                "LOGOUT_REDIRECT_URL", self.appbuilder.get_url_for_index
+            )
+        )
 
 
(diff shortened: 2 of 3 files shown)


heul:
I know it's a stale and closed issue, but this seems like a simple enough addition. @dpgaspar , is there a chance of still merging it?

kaojunsong:
@dpgaspar Hi Daniel, we also met the logout issue: https://github.com/dpgaspar/Flask-AppBuilder/issues/1110, and I think this MR can solve this problem, could you please help to merge the MR?

donatas-jocas:
I've also run into same issue mentioned by @kaojunsong using Keycloak SSO provider for Airflow - have verified in my local env that suggested changes in this PR solves the problem. @dpgaspar would it be possible to merge this PR for the upcoming release, even though it is marked as stale and closed? Happy to contribute making this happen if you have suggestions

jeverling:
I have merged latest master (`f0608fb`) into the branch to bring it up-to-date, but Github doesn't seem to pick that commit up automatically. Should I maybe create a new PR?

PR #1164, opened 2019-10-17
[security] New, state reason for LDAP login failure

CameronNemo:
This aids in security-related log analysis, because the failure and
reason for failure are clearly encapsulated into a single log line.

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -825,6 +825,7 @@ class BaseSecurityManager(AbstractSecurityManager):
                     # username = DN from search
                     username = user[0][0]
                 else:
+                    log.debug("LDAP bind failure: user not found")
                     return False
             log.debug("LDAP bind with: {0} {1}".format(username, "XXXXXX"))
             if self.auth_ldap_username_format:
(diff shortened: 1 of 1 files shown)


flask_appbuilder/security/manager.py
```diff
@@ -825,6 +825,7 @@ def _bind_ldap(self, ldap, con, username, password):
                     # username = DN from search
                     username = user[0][0]
                 else:
+                    log.info("LDAP bind failure: user not found")
```
dpgaspar: Let's make this debug also

flask_appbuilder/security/manager.py
```diff
@@ -835,6 +836,7 @@ def _bind_ldap(self, ldap, con, username, password):
             log.debug("LDAP bind OK: {0}".format(username))
             return True
         except ldap.INVALID_CREDENTIALS:
+            log.info("LDAP bind failure: invalid credentials")
```
dpgaspar: Could be unnecessary at this point since at we have the following:
https://github.com/dpgaspar/Flask-AppBuilder/blob/b4d0db9afa9639f86b201231584e7107810fc526/flask_appbuilder/security/manager.py#L903

and btw, this log at 903 should be a warning, can I ask you to change that also?

CameronNemo: That log will be emitted in both cases (user not found & password incorrect). I am specifically interested in a log message that is only emitted in the "user found, password incorrect" case.

flask_appbuilder/security/manager.py
CameronNemo: done

CameronNemo:
Bump. responded with why I think some of those changes do not make sense, adjusted the logging severity.

PR #1174, opened 2019-10-30
New, 1172 support jsonapi

### LINKED ISSUE #1172: Enhancement: Support forJSON:API-compliant requests and responses
Flask-AppBuilder provides some helpers to create CRUD API endpoints directly from a model. Which is great! Unfortunately there seem to be little to no options for customizing the request and response format, which means that I would have to write my own endpoint code from scratch for each individual HTTP verb if I want to accept requests and send responses that conform to e.g. the JSON:API standard. It would be great if there was some option that conditions API endpoints to expect requests and produce responses in the JSON:API format instead of what looks to be Flask-AppBuilder's own format.

zeusttu:
This PR adds support for accepting JSON:API-compliant requests and producing JSON:API-compliant responses when the `Accept` header is set to `application/vnd.api+json`. Ideally we would want to separate the expected request format and the produced response format because the HTTP standard governs them with respectively the `Content-Type` and `Accept` headers, but because they use the same Marshmallow schema and are nearly impossible to decouple, I used `Accept` for both.

In order to implement this change, I had to proxy some function calls with a defaultdict, and key them with a value that is not passed to the function in question, but is still available to it through a side-effect and influences its result. I wrote a pretty long comment in the code to explain this, but it may still prove to be a source of confusion, so do not hesitate to tell me if you have suggestions that would help further clarify this.

Finally, the diff of this pull request may be a bit overwhelming if you view it all at once, but if you review this PR on a commit-by-commit basis then every individual commit should have an easy-to-understand scope.

I did some API calls to test the new functionality as well as the old functionality, and it seemed to work, but my tests have been far from extensive. I also have not taken a look at how to run the unit tests yet, which also means I have not run any tests, nor have I added any tests for the new functionality.

Finally, this PR contains the basis to make the API mostly JSON:API-compliant, but a lot of error responses are still non-conformant to the standard, some HTTP response codes may or may not be nonconformant, and the OpenAPI spec bits I added do not specify any request/response body elements beyond what is included in `data`. I also have not tested with any models that include relationships, nor have I paid any attention to relationship-related functionality.

I feel that the scope of this PR would grow waaay too big if I aimed for full JSON:API compliance from the get-go, so probably we will have to create smaller issues for any non-compliant things that we find, and not pull them in scope of this PR. More extensive testing (especially for "old-style" (non-JSON:API) requests in more complex apps than the Hello World one I tested with) is still advisable before merging though, even if just to check that I did not break anything that worked before. Because of this, I will mark the PR as a draft for now, even though it very well may turn out to be ready.

Closes #1172.

### DIFF
--- a/flask_appbuilder/api/__init__.py
+++ b/flask_appbuilder/api/__init__.py
@@ -2,6 +2,7 @@ import functools
 import logging
 import re
 import traceback
+from collections import defaultdict
 
 from apispec import yaml_utils
 from flask import Blueprint, current_app, jsonify, make_response, request
@@ -16,6 +17,7 @@ import yaml
 
 from .convert import Model2SchemaConverter
 from .schemas import get_info_schema, get_item_schema, get_list_schema
+from .util import requested_type
 from .._compat import as_unicode
 from ..const import (
     API_ADD_COLUMNS_RES_KEY,
@@ -856,22 +858,22 @@ class ModelRestApi(BaseModelApi):
                     'gender': ('name', 'asc')
                 }
     """
-    list_model_schema = None
+    list_model_schema = defaultdict(lambda: None)
     """
         Override to provide your own marshmallow Schema
         for JSON to SQLA dumps
     """
-    add_model_schema = None
+    add_model_schema = defaultdict(lambda: None)
     """
         Override to provide your own marshmallow Schema
         for JSON to SQLA dumps
     """
-    edit_model_schema = None
+    edit_model_schema = defaultdict(lambda: None)
     """
         Override to provide your own marshmallow Schema
         for JSON to SQLA dumps
     """
-    show_model_schema = None
+    show_model_schema = defaultdict(lambda: None)
     """
         Override to provide your own marshmallow Schema
         for JSON to SQLA dumps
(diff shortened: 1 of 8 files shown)


zeusttu:
I haven't had time to resolve issues so far, but I might today. I'm starting by rebasing on your latest master, I'm about to force pushing to my feature branch. This force push does not include any new changes beyond this rebase.

zeusttu:
I see that `tox.ini` is configured to only run `flake8` and the nonexistent `tests` environment, but not any environment that actually runs the unit tests. When I configure it to run the `sqlite-api` environment, I find that most tests say `FAIL` or `ERROR`, but so do they on the `master` branch. @dpgaspar Are these tests expected to fail or should they pass? Is it possible that this is due to limitations of SQLite and that I have to create a Postgres or MySQL database to properly run the tests?

zeusttu:
Known remaining issues (if it says "(out of scope)", that means I would rather create a new issue for it than fix it as part of this PR):

- [ ] (out of scope) Error responses do not yet comply with the JSON:API format.
- [ ] (out of scope) Relationships are handled incorrectly in JSON:API responses, embedding the related object within `attributes`. Desired behaviour: put a reference in `relationships` and optionally include the related object in the `included` key.
- [ ] No new unit tests have been added yet. I will look if it makes sense to do so, and if so, I think I still want to make them part of this PR.

PR #1187, opened 2019-11-10
Fix, JWT refresh endpoint should return new access_token

alexandrezia:
When calling the JWT refresh endpoint (/security/refresh) using the refresh_token it should return a new access_token
But it's returning a new refresh_token
However looking at the FAB source code I think it is indeed returning a new access_token but naming it as refresh_token


dpgaspar:
Thank you (once more) @alexandrezia

PR #1193, opened 2019-11-19
[security] Fix, authentication error when using oracle dialect

rpepato:
This commit applies for oralce dialect the same query strategy already implemeted for MSSQL in
security manager when checking for authentication/permissions, since a
query like "SELECT EXISTS(1) FROM DUAL" in oracle databases would raise
an ORA-00936: missing expression error (observed from version 18c on)

Steps to reproduce the error:
1 - Configure a connection through the oracle dialect to an oracle instance (>=18c)
2 - Create an admin user through flask fab create-admin
3 - Login into the application

For Fixes:

For new features:

### DIFF
--- a/flask_appbuilder/security/sqla/manager.py
+++ b/flask_appbuilder/security/sqla/manager.py
@@ -311,8 +311,8 @@ class SecurityManager(BaseSecurityManager):
             )
             .exists()
         )
-        # Special case for MSSQL (works on PG and MySQL > 8)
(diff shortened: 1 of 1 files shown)


dpgaspar:
Hi @rpepato,

Will this fix work also on Oracle versions < 18c?

rpepato:
Tested for compatibility from Oracle 12c on and worked as expected

PR #1210, opened 2019-12-04
[view] New, non-attachment download endpoint

CameronNemo:
This provides an endpoint like the download endpoint, but more suitable for images that can be viewed standalone in a browser tab without being downloaded.

I used the same permission name because the content is the same.
Only the metadata of the response differ.


CameronNemo:
bump

CameronNemo:
Still think this is an important tool to display images.

PR #1215, opened 2019-12-09
redirect to next url after login if next params exist

feiming:

For Fixes:

For new features:

### DIFF
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -1,6 +1,7 @@
 import datetime
 import logging
 import re
+import urllib.parse
 
 from flask import abort, current_app, flash, g, redirect, request, session, url_for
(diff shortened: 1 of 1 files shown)


dpgaspar:
Thank you, but can you add a bit more context here, motivation (problem you are trying to solve), how does your change solve the problem. Also add tests

CameronNemo:
Currently, after login the user gets punted to the homepage / index. Regardless of where they intended to go prior to having to login. This change makes it so that the `next` parameter is actually used to redirect users who have just logged in. The `is_safe_url()` method (which is actually just a function -- no self usage) provides a preventative measure for phishing attempts, e.g. `https://myinternalapp.example.co/login/?next=<externalurl>`.

dpgaspar (commented):
Left a couple of comments

PR #1222, opened 2019-12-18
Note that 2.0.0 requires Python 3.6+

ashb:
The requirement is explicitly added in 2.1.12, but `f"${strings}"` have been
in use since e0f0e06


dpgaspar:
Top! thank you @ashb

PR #1239, opened 2020-01-15
[securitymanager] make userstatschartview optional

mistercrunch:
@dpgaspar let me know what you think, I was looking for a way to disable this view in Superset


dpgaspar:
tested it on SupersetSecurityManager and looks good `userstatschartview = None`

PR #1242, opened 2020-01-15
[security] Set JWT user on flask g

dpgaspar:
Set JWT auth user load to be set on flask `g`


CameronNemo:
You may want to test the AuditMixin with the API. I was having issues with it before so just opted not to use the mixin. This looks like it could fix the problem I noticed.

dpgaspar:
Great @CameronNemo, thanks for the feedback. Still not happy with this solution, but I'll push this forward

PR #1243, opened 2020-01-16
Fix, do not panic on unimplemented ordering

CameronNemo:
Sometimes a column cannot be ordered, and the app panics. This gives a friendly warning and renders the list without the ordering applied.


CameronNemo:
Wish I could remember what triggered these panics. IIRC it was the use of formatters_columns or maybe the renders decorator?

Issue #1114, opened 2019-09-10
Uploads pain points


CameronNemo:
If you'd like to report a bug in Flask-Appbuilder, fill out the template below. Provide
any extra information that may be useful

### Environment

Flask-Appbuilder version: 2.1.13

pip freeze output:

```
alembic==1.1.0
apispec==1.3.3
attrs==19.1.0
Babel==2.7.0
certifi==2019.3.9
chardet==3.0.4
Click==7.0
colorama==0.4.1
defusedxml==0.6.0
Flask==1.1.1
Flask-AppBuilder==2.1.13
Flask-Babel==0.12.2
Flask-JWT-Extended==3.22.0
Flask-Login==0.4.1
Flask-Migrate==2.5.2
Flask-OpenID==1.2.5
Flask-SQLAlchemy==2.4.0
Flask-WTF==0.14.2
idna==2.8
itsdangerous==1.1.0
Jinja2==2.10.1
jsonschema==3.0.2
Mako==1.1.0
MarkupSafe==1.1.1
marshmallow==2.19.5
marshmallow-enum==1.5.1
marshmallow-sqlalchemy==0.19.0
Pillow==6.1.0
prison==0.1.2
pyasn1==0.4.7
pyasn1-modules==0.2.6
PyJWT==1.7.1
pyrsistent==0.15.4
python-dateutil==2.8.0
python-editor==1.0.4
python-ldap==3.2.0
python3-openid==3.1.0
pytz==2019.2
PyYAML==5.1.2
requests==2.22.0
six==1.12.0
SQLAlchemy==1.3.8
SQLAlchemy-Utils==0.34.2
urllib3==1.25.3
uWSGI==2.0.18
Werkzeug==0.15.6
WTForms==2.2.1
```

### Describe the expected results

1. When using a non-existent directory for the uploads folder, the folder should be created with the executable bits set. Otherwise, flask-appbuilder will become unable to save files in said folder. I am having to manually set those permissions, even though I theoretically should not need to.

2. When specifying the path for the uploads folder without a trailing slash, flask-appbuilder fails to correctly provide the URL for files. Perhaps a strategic call to os.path.join could alleviate this issue.

```
sep 10 09:54:10 pgndb.tus.cenic.org docker[7449]: 2019-09-10 16:54:10,449:ERROR:app:Exception on /attachmentview/download/6b91ba4a-d3e9-11e9-a0a7-0242ac110003_sep_cat5172.jpg [GET]
[...]

PR #1281, opened 2020-02-15
[mvc] Fix, use os.path.join for download

CameronNemo:
#1114


dpgaspar:
Thank you @CameronNemo !

PR #1330, opened 2020-03-27
New - example to integrate Dash by Plotly in FAB

giordyb:
Example to integrate Dash by Plotly in FAB protected via login.


dpgaspar:
Nice! looks good

Issue #1182, opened 2019-11-07
Hardcoding the class name in login_oauth.html and register_oauth.html


snivas:
There is an issue when we create a custom class for handling AuthOAuthView. The issue is caused by below lines,

> -var baseLoginUrl = "{{url_for('AuthOAuthView.login')}}";
> -var baseRegisterUrl = "{{url_for('AuthOAuthView.login')}}";

Should we must give "AuthOAuthView.login" there? suppose if we have a custom class that implements AuthOAuthView with name CustomOAuthView then flask is throwing error. Shall we change like below?

> 
> +var baseLoginUrl = "{{appbuilder.get_url_for_login}}";
> +var baseRegisterUrl = "{{appbuilder.get_url_for_registeruser}}";

Likely, 

why we need url parameter in "function set_openid(**url**, pr)". As it is not used, shall we remove it?

dpgaspar:
Seems right, can you open a PR?

snivas:
Could you give permission for me to commit?

PR #1331, opened 2020-03-27
Fix, hardcoded url in oauth template, fix for #1182

giordyb:

For Fixes:

fix hardcoded url in oauth template see #1182


dpgaspar:
Looks good, I'l test it first

PR #1324, opened 2020-03-21
[i18n] New, Add italian translation

nbenaglia:
Add italian translation

PR #1332, opened 2020-03-27
New - Italian translation

giordyb:
Italian translation - mine is a bit more "formal" than #1324


dpgaspar:
oh... just merged #1324, can you rebase and just address the necessary changes?

giordyb:
It looks like we based the translation on different templates so it's a bit complicated to just address the differences, I would have to review all of them because they are in a different order...May I suggest that you go ahead and disregard my pull request? The translation you already merged looks fine to me anyways.

dpgaspar:
Sure, sorry for the this and thanks anyway

Issue #1367, opened 2020-05-05
name column resizing in ab_view_menu table 


guptakumartanuj:
While working with apache-airflow when we enable rbac, all the tables related to flask-ui are added into the metadata. On that note, system tries to insert the rows in ab_view_menu table with respect to all the available DAGs. The length of the dag_id is 250 chars but in ab_view_menu the aligned column's (name) length is 100 chars and in turn value is truncated. Thereafter, it throws the Integrity Error when tries to insert the same value of name because of the truncation. And finally database goes down.

So, to improve this, we need to keep in sync with the ab_view_menu table. For that, I need to change the size of the column from 100 to 250.

guptakumartanuj:
@dpgaspar Can you please look into this ?

PR #1368, opened 2020-05-05
Fix, name column resizing in ab_view_menu table #1367

guptakumartanuj:
While working with apache-airflow when we enable rbac, all the tables related to flask-ui are added into the metadata. On that note, system tries to insert the rows in ab_view_menu table with respect to all the available DAGs. The length of the dag_id is 250 chars but in ab_view_menu the aligned column's (name) length is 100 chars and in turn value is truncated. Thereafter, it throws the Integrity Error when tries to insert the same value of name because of the truncation. And finally database goes down.

So, to improve this, we need to keep in sync with the ab_view_menu table. For that, I need to change the size of the column from 100 to 250.

https://github.com/dpgaspar/Flask-AppBuilder/issues/1367

### DIFF
--- a/flask_appbuilder/security/sqla/models.py
+++ b/flask_appbuilder/security/sqla/models.py
@@ -33,7 +33,7 @@ class Permission(Model):
 class ViewMenu(Model):
     __tablename__ = "ab_view_menu"
     id = Column(Integer, Sequence("ab_view_menu_id_seq"), primary_key=True)
-    name = Column(String(100), unique=True, nullable=False)
+    name = Column(String(250), unique=True, nullable=False)
 
     def __eq__(self, other):
         return (isinstance(other, self.__class__)) and (self.name == other.name)


guptakumartanuj:
@dpgaspar Can you please review this ?

guptakumartanuj:
@dpgaspar Not sure if u have got the chance to check this out of your busy schedule ? Please check the same.

guptakumartanuj:
@dpgaspar Thanks for reviewing the PR. Moreover, [Airflow](https://github.com/apache/airflow/tree/master/airflow/migrations/versions) doesn't keep any alembic version changes regarding Flask-AppBuilder. All the rbac related tables are imported to apche-airflow metadata as it is. Moreover, please let me know if any I need update any alembic version change in this repo.

guptakumartanuj:
@dpgaspar Can you please merge this if you don't have any other ask in this ?

ashb:
@dpgaspar Do you think we (Airflow) should start tracking changes (and "creation") in our Alembic migrations? How often to things change in the tables?

dpgaspar:
@ashb That's a very good question. Last model change was years ago. 
- One way to go could be to include this change on Airflow alembic since it's critical for you.
- I can include a sort of schema migration on FAB, triggered by cli. But I'm keeping FAB agnostic to the mig tool people use. and this would always be far from perfect

Happy to discuss this further!

ashb:
This is an old bug that's been around since about Airflow 1.10.2(? .5?) when we introduced per-dag permissions, it's just not _that_ commonly hit so we haven't noticed it before.

guptakumartanuj:
@dpgaspar  Can this be merged ?

guptakumartanuj:
@dpgaspar Can we let me know what needs to be done  to merge this PR ?

dpgaspar:
Merging, planning to include this one on FAB 3.0.0, that will probably include marshmallow 3 also. FAB 3, will need a lot of testing and marshmallow 3 support is not ready yet.

ashb:
@dpgaspar Do you think it is a good idea for Airflow to have a migration that does a bunch of `CREATE TABLE IF NOT EXIST` for the "current" definition of the `ab_*` tables?

(`IF NOT EXIST` is to cope with new installs with an empty DB, and upgrading where we have been relying on the auto-creation of the tables by FAB.)

PR #1379, opened 2020-05-26
Syntax error when showing picture

HenkVanHoek:
The plus sign is not allowed. Causing an error when showing the product.

For Fixes:

For new features:


HenkVanHoek:
As requested, the f'function is used now.

PR #1386, opened 2020-06-04
New, list_query option on ModelRestApi

willbarrett:
This PR introduces a new option on the `ModelRestApi` called `list_query`. The option takes a callable that accepts a database session and returns a base SqlAlchemy Query object to use for the list view.

Check out the added tests for example usage. When paired with the `list_model_schema` option, this new option allows for counts, rollups, and other more sophisticated behavior on the list endpoints than is possible currently. It also allows for complete control over how FAB loads data on list views (specifying join options and the like).

@dpgaspar your recommendations on how to improve and document this would be greatly appreciated! Thanks for your assistance so far!


flask_appbuilder/tests/test_api.py
```diff
@@ -1067,6 +1104,29 @@ def test_get_list_base_order(self):
...
+        """
+            REST Api: Test get list with a base query
+            defined on the API for the list endpoint
+        """
+
+        client = self.app.test_client()
+        token = self.login(client, USERNAME_ADMIN, PASSWORD_ADMIN)
+
+        # test string order asc
+        rv = self.auth_client_get(client, token, "api/v1/model1apibasequery/")
```
dpgaspar, line 1117: We should make sure that the filters,order by and pagination still work has expected

flask_appbuilder/models/sqla/interface.py
```diff
@@ -200,10 +201,15 @@ def query(
...
             the current page size
+        :param base_query
+            the base SQLAlchemy Query to run to build the list view
         """
-        query = self.session.query(self.obj)
+        if base_query:
+            query = base_query(self.session)
+        else:
+            query = self.session.query(self.obj)
+            query = self._query_select_options(query, select_columns)
```
dpgaspar, line 211: Note: query_count is still counting the obj rows, it should count the passed query

willbarrett: Good catch!

flask_appbuilder/tests/test_api.py
willbarrett: I'll add some tests!

PR #1334, opened 2020-03-27
[dependencies] Fix, support marshmallow 3

dpgaspar:
Support marshmallow 3

This is a breaking change!

- API is strict by default, so POST, PUT with undeclared fields will HTTP 422
- New support for functions/methods (callables) has model properties

PR #1392, opened 2020-06-08
SQLAlchemyAutoSchema needs marshmallow-sqlalchemy>=0.22.0

ashb:
#1334 started using this class, but didn't update the requirements in setup.py, so it would fail in an existing environment


dpgaspar:
Thank you!

PR #1406, opened 2020-06-18
Fix: fixed typos and improved bit of the German translation

BalthasarSchachtner:
Minor fixes for the German translation I noted when using the app.


dpgaspar:
Thank you @BalthasarSchachtner

BalthasarSchachtner:
Dear @dpgaspar, I have pulled the newest version and just realized, that my pull request did not introduce the changes as I expected, it seems like the translation file needs to be recompiled and checked in.
I'm not familiar with the corresponding tools, could you please point me to them or rerun the compilation yourself?
Thanks,
Balthasar

PR #1408, opened 2020-06-21
Fix, action confirmation disabling

mikitex70:
Fixed action disabling.

When using action with `confirmation=None`, a confirmation dialog still popups, with `None` message.
This is because the `None` Python value is translated in a `None` Javascript string, which is different from a `null` value and then `!!confirmation` is always `true`.

The fix checks for a `None` string value and convert it to a `false` Javascript value.

### DIFF
--- a/flask_appbuilder/templates/appbuilder/general/lib.html
+++ b/flask_appbuilder/templates/appbuilder/general/lib.html
@@ -1,10 +1,10 @@
 {% macro render_additional_links(additional_links, pk) %}
-	{% if additional_links %}
-		{% for item in additional_links %}
-			<a href="{{item.href}}?pk={{pk}}" class="btn btn-sm btn-primary"><span class="glyphicon glyphicon"></span>
-			{{_(item.label)}}</a>
-		{% endfor %}
-	{% endif %}
+    {% if additional_links %}
+        {% for item in additional_links %}
+            <a href="{{item.href}}?pk={{pk}}" class="btn btn-sm btn-primary"><span class="glyphicon glyphicon"></span>
+            {{_(item.label)}}</a>
+        {% endfor %}
+    {% endif %}
 {% endmacro %}
 
 {% macro render_action_links(actions, pk, modelview_name) %}
@@ -59,7 +59,7 @@
     {{_('Actions')}}<span class="caret"></span>
     </button>
     <ul class="dropdown-menu" role="menu">
-    	{% for action_key in actions %}
+        {% for action_key in actions %}
             {% set action = actions.get(action_key) %}
                 <li>
                     <a href="javascript:void(0)"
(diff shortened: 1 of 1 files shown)


mik-laj:
https://github.com/dpgaspar/Flask-AppBuilder/blob/ce6caa255a6b0e279d1e7dc8b2e51bc34cf60e94/flask_appbuilder/templates/appbuilder/general/lib.html#L78
```diff
				$('.{{action.name}}_menu_item').on('click', function(){
-				    return modelActions.execute_multiple('{{action.name}}','{{action.confirmation}}');
+                                   {% if confirmation %}
+				        return modelActions.execute_multiple('{{action.name}}','{{action.confirmation}}');
+                                   {% else %}
+				        return modelActions.execute_multiple('{{action.name}}', false);
+                                   {% endif %}
				})
```

mikitex70:
> https://github.com/dpgaspar/Flask-AppBuilder/blob/ce6caa255a6b0e279d1e7dc8b2e51bc34cf60e94/flask_appbuilder/templates/appbuilder/general/lib.html#L78
> 
> ```diff
> 				$('.{{action.name}}_menu_item').on('click', function(){
> -				    return modelActions.execute_multiple('{{action.name}}','{{action.confirmation}}');
> +                                   {% if confirmation %}
> +				        return modelActions.execute_multiple('{{action.name}}','{{action.confirmation}}');
> +                                   {% else %}
> +				        return modelActions.execute_multiple('{{action.name}}', false);
> +                                   {% endif %}
> 				})
> ```

I don't understand: there was already a pull request for the issue, or this is a suggestion for a better way to fix it?

mik-laj:
I think that this problem should be solved before it appears.  what do you think to fix it this way?

mikitex70:
> I think that this problem should be solved before it appears. what do you think to fix it this way?

I like your solution, a lot better than mine.

mik-laj:
Can you update your PR and include my solution?

flask_appbuilder/templates/appbuilder/general/lib.html
```diff
@@ -75,7 +75,11 @@
 		{% for action_key in actions %}
         {% set action = actions.get(action_key) %}
 				$('.{{action.name}}_menu_item').on('click', function(){
-				    return modelActions.execute_multiple('{{action.name}}','{{action.confirmation}}');
+                    {% if confirmation %}
```
dpgaspar, line 78: `confirmation` comes from where?

dpgaspar, line 78: indentation looks really bad here, could you take the chance and fix it also?

mikitex70: I don't know, I supposed suggestion from @mik-laj was correct.
Let me do some check.

mik-laj: It should be action.confirmation

mikitex70: In the file there was tabs and spaces before my editing. I've tried to keep indentation using tabs...
I will replace all tabs with spaces to keep indentation coherent with any editor.

dpgaspar: Thank you!

PR #1410, opened 2020-06-22
New, Dynamic user registration role

szczeles:
This PR brings dynamic user role on registration based on OAuth userinfo.

The current approach sets static `AUTH_USER_REGISTRATION_ROLE` for all users if self-registration is enabled. In our case, with OAuth authentication on GSuite domain (with FAB being auth engine for Airflow), we'd like to set team members as Admins and every other employee as read-only user ("Viewer"). 

This PR implements method that is used in Grafana ([docs](https://grafana.com/docs/grafana/latest/auth/generic-oauth/#jmespath-examples)), it uses [JMESPath](https://jmespath.org/) to resolve the role from userinfo retrieved from OAuth. It's quite flexible, as some OAuth providers return only username (like Google does), but others, like Okta, can return groups, so they can be utilized by the dynamic expression.

If needed, I'm more than happy to extend the docs by adding more examples of JMESPath.

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -349,7 +349,7 @@ class BaseSecurityManager(AbstractSecurityManager):
         return self.appbuilder.get_app.config["AUTH_USER_REGISTRATION_ROLE"]
 
     @property
-    def auth_user_registration_role_jmespath(self):
+    def auth_user_registration_role_jmespath(self) -> str:
         return self.appbuilder.get_app.config["AUTH_USER_REGISTRATION_ROLE_JMESPATH"]
 
     @property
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -22,6 +22,7 @@ flask==1.1.1
 idna==2.9                 # via email-validator
 itsdangerous==1.1.0       # via flask
 jinja2==2.10.1            # via flask, flask-babel
+jmespath==0.9.5
```
dpgaspar, line 25: If it's not on `setup.py` it can't be here, this file is auto generated by pip-compile. We could add it to requirements-dev.txt if we decide it's not a root dependency

flask_appbuilder/security/manager.py
```diff
@@ -347,6 +348,10 @@ def auth_user_registration(self):
     def auth_user_registration_role(self):
         return self.appbuilder.get_app.config["AUTH_USER_REGISTRATION_ROLE"]
 
+    @property
+    def auth_user_registration_role_jmespath(self):
```
dpgaspar, line 352: nit: I'm starting to add type annotations, at least on every new PR.

`def auth_user_registration_role_jmespath(self) -> str:`

requirements.txt
szczeles: Thanks! I think it's not a root dependency. Added to extras.

szczeles:
@dpgaspar Thanks for the review! I've moved the dependency to setup's extras and added a few lines of documentation with examples

PR #1413, opened 2020-06-24
Fix, security issue with built in roles

john-bodley:
@dpgaspar this PR fixes an issue we noticed with the `FAB_ROLES` feature as it seems that the `FAB_ROLES` configuration was acting as a replacement to the backend terms of access checks rather than a precursor to the backend checks as documented in the code. 

The problem was if one had a role defined in the `FAB_ROLES` but the permission name didn't exist then the roles was never added to the set which should be checked in the backend.

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -1091,17 +1091,16 @@ class BaseSecurityManager(AbstractSecurityManager):
         self, user: object, permission_name: str, view_name: str
     ) -> bool:
         roles = user.roles
-        db_role_ids = list()
         # First check against builtin (statically configured) roles
         # because no database query is needed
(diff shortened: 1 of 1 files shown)


john-bodley:
@dgparser I see. Per the code comment 

> Then check against database-stored roles

I wrongly interpreted the logic as the database was a fallback if the permission/role isn’t defined  in the `FAB_ROLES` as opposed to just the role.

dpgaspar:
Reread that and your right, it's misleading, maybe it should be something like:

> If it's not a builtin role check against database stored roles

?

john-bodley:
@dpgaspar I think that's clearer. Initially it wasn't apparent to whether the `FAB_ROLES` were a replacement or a complement.

PR #1457, opened 2020-08-26
Update baselib.html

pasqwal:
[macro navbar_block] : Make the App Icon rendering exactly the same as it is defined in navbar.html (which is extended by the baselayout)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


pasqwal (commented):
This is a minor to make the App Icon rendering exactly the same as it is defined in navbar.html (which is extended by the baselayout)

PR #1465, opened 2020-09-02
fix: #1453

### LINKED ISSUE #1453: Invalid URLs for OpenApi + SwaggerView with relativ url root
If you'd like to report a bug in Flask-Appbuilder, fill out the template below. Provide
any extra information that may be useful

### Environment

Flask-Appbuilder version:
>3

pip freeze output:
unimportant

### Describe the expected results
When using openapi and swagger view with a relativ url root (e.g. localhost/foo/bar/api/v1) urls are not correct.
A possible fix could be the following:

in api/manager.py 
```
 servers=[{"url": url_for("OpenApi.get", version=version)[:-9]}],
 openapi_uri=url_for(OpenApi.__name__ + "." + OpenApi.get.__name__, version=version),
```

Tell us what should happen.

```python
Paste a minimal example that causes the problem.
```

### Describe the actual results

Tell us what happens instead.

```pytb
Paste the full traceback if there was an exception.
```

lnnwvr:
Fixes #1453


dpgaspar (changes requested):
Thank you. Looks good, but you need to run `black` and `flake8`

lnnwvr:
finally...
fyi: i had some trouble with flake8 and also psycopg2-binary(which should be upgraged) and python 3.8

flask_appbuilder/api/manager.py
```diff
@@ -58,22 +58,23 @@ def _create_api_spec(version):
             openapi_version="3.0.2",
             info=dict(description=current_app.appbuilder.app_name),
             plugins=[MarshmallowPlugin()],
-            servers=[{"url": "/api/{}".format(version)}],
+            servers=[{"url": url_for("OpenApi.get", version=version)[:-9]}],
```
dpgaspar, line 61: Sorry, what this for `[:-9]` ?

dpgaspar (changes requested):

flask_appbuilder/api/manager.py
lnnwvr: url_for("OpenApi.get", version=version) returns url_root + /api/v1/_openapi but we just need url_root + /api/v1

mjpieters: I’d use `.rpartition("/")[0]`, which splits the string at the last slash and returns everything before that point.

Isn’t there a better source for the API version root URL however?

dpgaspar: The class `OpenApi` itself holds the base and the version is passed to this method. The API version can vary on each `BaseApi` class so the source is at that level

lnnwvr: so what do we change?

mjpieters: Make it

```python
servers=[{"url": url_for("OpenApi.get", version=version).rpartition("/")[0]}],
```

although I'd pull out the expression into a separate variable:

```python
    @staticmethod
    def _create_api_spec(version):
        # The spec produced here applies to the base of this view:
        api_url = url_for("OpenApi.get", version=version).rpartition("/")[0]
        return APISpec(
            title=current_app.appbuilder.app_name,
            version=version,
            openapi_version="3.0.2",
            info=dict(description=current_app.appbuilder.app_name),
            plugins=[MarshmallowPlugin()],
            servers=[{"url": api_url}],
        )
```

lnnwvr: i made the changes, how do i merge the change it into the pull request .. sometimes i dont understand github

lnnwvr:
Its still a issue isnt it @dpgaspar ?

PR #1481, opened 2020-10-06
Returning None is a better approach. 

mfyuce:
- The method normall returns a User. In case of an error a boolean. Thus changing type may cause other unpredicted behaviour
- in changed lines, user may already be none. A method called with it, will divert the problem to another place.

### Description
When ever a user cannot be add, edited function returnes a boolean value. And do not throw. Because of that, the code after that can get problems seeking the root cause of the exception

Only returned None instead of False which is not the same type with a user.
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ X] Is Auth, RBAC security related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -1053,11 +1053,11 @@ class BaseSecurityManager(AbstractSecurityManager):
                 email=userinfo.get("email", ""),
                 role=self.find_role(role_name),
             )
-            if not user:
+            if user is None:
                 log.error("Error creating a new OAuth user %s" % userinfo["username"])
                 return None
-        if user:
-            self.update_user_auth_stat(user)
+
+        self.update_user_auth_stat(user)
         return user
 
     """
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -955,7 +955,8 @@ class BaseSecurityManager(AbstractSecurityManager):
                             role=self.find_role(self.auth_user_registration_role),
                         )
 
-                self.update_user_auth_stat(user)
+                if user:
+                    self.update_user_auth_stat(user)
                 return user
 
             except ldap.LDAPError as e:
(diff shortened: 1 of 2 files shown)


flask_appbuilder/security/manager.py
```diff
@@ -1054,7 +1056,8 @@ def auth_user_oauth(self, userinfo):
             if not user:
```
dpgaspar, line 1056: Now we should `if user is None`

flask_appbuilder/security/manager.py
```diff
@@ -1054,7 +1056,8 @@ def auth_user_oauth(self, userinfo):
...
                 log.error("Error creating a new OAuth user %s" % userinfo["username"])
                 return None
-        self.update_user_auth_stat(user)
+        if user:
```
dpgaspar, line 1059: We have already made sure the user exists here, or not?

flask_appbuilder/security/sqla/manager.py
```diff
@@ -214,7 +214,7 @@ def add_user(
         except Exception as e:
             log.error(c.LOGMSG_ERR_SEC_ADD_USER.format(str(e)))
             self.get_session.rollback()
-            return False
+            return None
```
dpgaspar, line 217: Agree! would be nice to add type annotations to add_user since we are here

dpgaspar (commented):
Looks good  but left a couple of comments

flask_appbuilder/security/manager.py
mfyuce: done

flask_appbuilder/security/manager.py
mfyuce: true

flask_appbuilder/security/sqla/manager.py
mfyuce: added as "->User" if you are OK with it.

mfyuce: removed "->User", as the  user model can be changed at run time

Issue #1402, opened 2020-06-16
allow_blank=True does not do anything for QuerySelectMultipleField warning


andreamancini-work:
Hi, 
I'm using your package from a pair of mouths and it is very good.
Sometimes in logs appears this warning, but I can't understand if it is caused by my wrong configuration or others.

```
/home/pi/.local/lib/python3.7/site-packages/flask_appbuilder/fields.py:182: UserWarning: allow_blank=True does not do anything for QuerySelectMultipleField.
  "allow_blank=True does not do anything for QuerySelectMultipleField."
```

Thanks in advance.

mapio:
The issue has been closed as stale, but the bug is still present and the pull request fixing it is stale (since unprocessed).

PR #1508, opened 2020-10-28
fix: Update forms.py

mapio:
Removed `allow_blank` kwarg set to `True` by default  in `QuerySelectMultipleField` instantiation, due to the warning for that value in such class.

### Description

When using a multiple choice field due to an enumeration in a model a warning is raised by `QuerySelectMultipleField` complaining about the `True` value of the `allow_blank` kwarg.

This patch just removes the kwarg from the instantiation of a `QuerySelectMultipleField`.

### ADDITIONAL INFORMATION

This eliminates the warning reported in #1402 and seems not to change the behaviour of the field in the form.


mapio:
It's a pity this is stale, the fix is still required.

PR #1509, opened 2020-10-30
fix: Primary Keys need not to be integer

mapio:
Primary keys in `views.py` were converted to ints using `int()´. This is an issue for models that have, for instance, strings as primary keys.

### DIFF
--- a/flask_appbuilder/views.py
+++ b/flask_appbuilder/views.py
@@ -2,7 +2,7 @@ import json
 import logging
 import os.path as op
 from typing import Set
-
+f
 from flask import (
     abort,
     flash,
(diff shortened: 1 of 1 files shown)


flask_appbuilder/views.py
```diff
@@ -432,7 +432,7 @@ def _get_related_column_data(self, col_name, filters):
         ret_list = list()
         for item in result:
             pk = rel_datamodel.get_pk_value(item)
-            ret_list.append({"id": int(pk), "text": str(item)})
+            ret_list.append({"id": pk, "text": str(item)})
```
mapio, line 435: Testing gitpod.

dpgaspar (approved):
LGTM, thanks for the fix

mapio:
Is there a reason why the patch that has been approved by @dpgaspar on Jan 4 has not yet been merged? Now it's stale, but for me the problem persist and I've to reapply the patch every time I upgrade the package.

PR #1542, opened 2021-01-08
fix: modal confirmation-text is wrong (ab_actions.js)

runoutnow:
### Description

**Problem** 
After you have done an `execute_single_delete` from `ab_actions.js` (which is used on lib.html) and hit cancel on that modal. The confirmation-text wil not update anymore for other modal-confirmations which make use of `execute_single_`

**Fix** 
`$("#modal-confirm .modal-body").text(confirmation); ` was missing in the `this.execute_single function` on `ab_actions.js`

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


R-Escher:
I believe the line should be changed from
`$("#modal-confirm .modal-body").text(confirmation);`
to
`$("#modal-confirm .modal-text").text(confirmation);`
because [this function here](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/static/appbuilder/js/ab_actions.js#L129) writes to the _.modal-text_, which is inside _.modal-body_.
Once you add text to the _.modal-body_, you can't see the text inside _.modal-text_ anymore (at least I couldn't in my browser).

PR #1590, opened 2021-03-21
fix: handle better missing Azure AD info when registering a new user

jonasjancarik:
### Description

Introduces logic for handling missing user meta information. Currently, if e.g. the first or last name are missing, registration fails with no clear indication of what went wrong.

This is based on a real issue we had with registering users with no first name or last name in their Azure AD user profile.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -599,13 +599,14 @@ class BaseSecurityManager(AbstractSecurityManager):
             log.debug(str(id_token))
             me = self._azure_jwt_token_parse(id_token)
             log.debug("Parse JWT token : {0}".format(me))
+            upn_split = me["upn"].split("@") if "upn" in me else []
             return {
-                "name": me["name"],
-                "email": me["upn"],
-                "first_name": me["given_name"],
-                "last_name": me["family_name"],
+                "name": me["name"] if "name" in me else "N/A",
+                "email": me["upn"] if "upn" in me else "N/A",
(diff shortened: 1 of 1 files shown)


dpgaspar:
@ jonasjancarik makes sense, but lint is failling

dpgaspar (changes requested):
@jonasjancarik makes sense, but lint is failing

jonasjancarik:
Yeah I'll try to get around to it and fix that.

I am also thinking that using just the first part of the email address (which the "upn" usually is - but it can also be set to something else) for username is not ideal - if you have people registering from different tenants, e.g. john@foo.com and john@bar.com, john wouldn't be a unique username.

So I will probably change that to using the whole email address (upn). (Currently the username will be based on the "oid" which is a long alphanumeric string - upn will be more readable.)

dpgaspar:
probably related to: https://github.com/dpgaspar/Flask-AppBuilder/pull/1640

thesuperzapper:
@jonasjancarik do you want to rebase and fix linting?

thesuperzapper:
@jonasjancarik is this still something you are working on?

jonasjancarik:
Hi, yes - sorry I haven't had a lot of time available for this. I was just checking what the linting issue was; I will try to rebase and push again

PR #1592, opened 2021-03-26
fix: remove pinned sqlalchemy and sqlalchemy-utils

dpgaspar:
### Description

Flask-SQLAlchemy released a new version that already supports SQLAlchemy 1.4: https://github.com/pallets/flask-sqlalchemy/blob/master/CHANGES.rst#version-250

Removing sqlalchemy-utils since it's not really used on FAB and it's not compatible for now:
https://github.com/kvesteri/sqlalchemy-utils/issues/505

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/models/sqla/interface.py
+++ b/flask_appbuilder/models/sqla/interface.py
@@ -13,7 +13,6 @@ from sqlalchemy.orm.session import Session as SessionBase
 from sqlalchemy.orm.util import AliasedClass
 from sqlalchemy.sql.elements import BinaryExpression
 from sqlalchemy.sql.sqltypes import TypeEngine
-from sqlalchemy_utils.types.uuid import UUIDType
 
 
 from . import filters, Model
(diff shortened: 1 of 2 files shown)


ashb:
@dpgaspar What was the thinking here? I'd _love_ to be able to use SQLA 1.4 in Airflow, but this dep is stopping us.

dpgaspar:
@ashb 

Yes and I would love to get rid of this hard pin for SQLAlchemy. What's stopping me is `sqlalchemy-utils` and the fact that superset depends on it's support on FAB also

dpgaspar:
Just checked and https://github.com/kvesteri/sqlalchemy-utils/blob/master/CHANGES.rst#0370-2021-04-12

So we are about to be unblocked

ashb:
Oh sweet! I looked this morning and saw nothing ❤️

PR #1595, opened 2021-04-01
fix calling datamodel.get_search_columns_list when not unnecessary // Fix #1594

### LINKED ISSUE #1594: Instantiate ModelRestApi with search_columns can failed for nothing
### Environment

Flask-Appbuilder version: 3.2.1 
Flask-SQLAlchemy       2.5.1
marshmallow-sqlalchemy 0.23.1
mysqlclient            1.4.2.post1
SQLAlchemy             1.3.23
SQLAlchemy-Utils       0.36.8
sqlparse               0.3.0

I have a restAPIDefinition 

```python
class MyRestApi(ModelRestApi):
    datamodel = SQLAInterface(MyModel)
    search_columns = ['id', 'name', 'context', 'owner_id']
```
and one of the properties of the model is 

```python
from sqlalchemy_utils import generic_relationship
owner_object = generic_relationship(owner_id, owner_type)
```

when It tries to initiate it after add_api it failed on line 841 due the use of generic_relationship
but It shouldn't call it anyway since self.search_columns is defined

image

Ofeknielsen:
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/api/__init__.py
+++ b/flask_appbuilder/api/__init__.py
@@ -838,8 +838,8 @@ class BaseModelApi(BaseApi):
         self._base_filters = self.datamodel.get_filters().add_filter_list(
             self.base_filters
         )
-        search_columns = self.datamodel.get_search_columns_list()
         if not self.search_columns:
+            search_columns = self.datamodel.get_search_columns_list()
             self.search_columns = [
                 x for x in search_columns if x not in self.search_exclude_columns
             ]


dpgaspar:
@Ofeknielsen the idea is that when not provided we default to allow for search/filter on all possible columns on the model

dpgaspar (changes requested):
This is a breaking change, and it's not clear what it actually fixes. Can you provide a better description please?

Issue #1581, opened 2021-03-05
LDAP authentication fails when the domain server response includes SearchResultReference(s)


DaleMittleman:
I'm working on setting up an instance of airflow 2.0.0.  Everything has been going smoothly.  I have the web app integrated with my company's Active Directory, users were able to login and were added to the default role upon first login.

Eager to plug into the new group --> role mappings feature (AUTH_ROLES_MAPPING) added in FAB 3.2.0, I upgraded FAB to 3.2.0.  After the upgrade, users could no longer login. 

I've sanitized some of the configuration and the log output below, but I believe it's still legible.

### Environment

Flask-Appbuilder version: 3.20
python-ldap version: 3.3.1 (I've also tested against every major version since and including 3.0.0)

relevant env variables:
```
AIRFLOW__LOGGING__FAB_LOGGING_LEVEL=DEBUG
```

webserver_config.py:
```
from flask_appbuilder.security.manager import AUTH_LDAP
import os

basedir = os.path.abspath(os.path.dirname(__file__))

WTF_CSRF_ENABLED = True

AUTH_TYPE = AUTH_LDAP
AUTH_ROLE_ADMIN = 'Admin'
AUTH_USER_REGISTRATION = True
AUTH_USER_REGISTRATION_ROLE = 'Viewer'

AUTH_LDAP_SERVER = '< santized >' 
AUTH_LDAP_SEARCH = '< santized >'

AUTH_LDAP_BIND_USER = '< santized >'
AUTH_LDAP_BIND_PASSWORD = os.getenv('AUTH_LDAP_BIND_PASSWORD')

AUTH_LDAP_UID_FIELD = 'sAMAccountName'
AUTH_LDAP_FIRSTNAME_FIELD = 'givenName'
AUTH_LDAP_LASTTNAME_FIELD = 'sn'
AUTH_LDAP_EMAIL_FIELD = 'mail'

AUTH_ROLES_MAPPING = {
    '< santized >': ["Admin"],
    '< santized >': ["User"],
}
AUTH_ROLES_SYNC_AT_LOGIN = True

AUTH_LDAP_GROUP_FIELD = "memberOf"

PERMANENT_SESSION_LIFETIME = 1800
```

### Describe the expected results

I log into the web app.  If this is a first time login, a user is created for me.  I'm added to the default role,  and the "Admin" and/or "User" roles if I am a member of the corresponding groups.

### Describe the actual results

[...]

PR #1602, opened 2021-04-07
fix: Ignore LDAP search referrals

fredthomsen:
### Description

LDAP referrals are completely broken anyway, and we've already set
`ldap.OPT_REFERRALS` to 0, so there is no reason to care about these.

Thus the presence of referrals alone should not cause authentication of
a user to fail due to multiple results being returned.

Fixes issue #1581.

If this change is not desired, and rather a documentation update to
indicate that setting up LDAP against Microsoft AD requires a
organizational unit to be specified in the LDAP search base, and I
happy to make that update.  I can add the necessary tests once I get
feedback if we want to move forward with this.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -878,10 +878,17 @@ class BaseSecurityManager(AbstractSecurityManager):
                 filter_str, request_fields, self.auth_ldap_search
             )
         )
-        search_result = con.search_s(
+        raw_search_result = con.search_s(
             self.auth_ldap_search, ldap.SCOPE_SUBTREE, filter_str, request_fields
         )
-        log.debug("LDAP search returned: {0}".format(search_result))
+        log.debug("LDAP search returned: {0}".format(raw_search_result))
+
+        # Remove any search referrals from results
+        search_result = [
+            (dn, attrs)
+            for dn, attrs in raw_search_result
+            if dn is not None and isinstance(attrs, dict)
+        ]
 
         # only continue if 0 or 1 results were returned
         if len(search_result) > 1:
--- a/flask_appbuilder/tests/_test_auth_ldap.py
+++ b/flask_appbuilder/tests/_test_auth_ldap.py
@@ -1,7 +1,7 @@
 import logging
 import os
 import unittest
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 from flask import Flask
 from flask_appbuilder import AppBuilder, SQLA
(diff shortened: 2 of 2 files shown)


flask_appbuilder/security/manager.py
```diff
@@ -878,9 +878,15 @@ def _search_ldap(self, ldap, con, username):
...
-            self.auth_ldap_search, ldap.SCOPE_SUBTREE, filter_str, request_fields
-        )
+
+        # filter out any referrals that get returned
+        search_result = [
+            (dn, attrs)
+            for dn, attrs in con.search_s(
+                self.auth_ldap_search, ldap.SCOPE_SUBTREE, filter_str, request_fields
+            )
+            if dn is not None
```
thesuperzapper, line 888: I'm pretty sure its a bad idea to just ignore any search reults with None for `dn`.

Are we 100% sure the only case this occurs in is SearchResultReference, because I doubt this.

Isn't there a better way to filter SearchResultReference instances explicitly.

thesuperzapper (changes requested):
I also think we should add some unit tests for this, but possibly that is overkill. 

BTW: you would need to add refs to [our LDAP test structure](https://github.com/dpgaspar/Flask-AppBuilder/blob/v3.2.2/flask_appbuilder/tests/_test_auth_ldap.py#L68), similar to how [dalibo does it](https://github.com/dalibo/ldap2pg/blob/5.5/tests/unit/test_manager.py#L10-L15), but do note that they use a different LDAP test library.

flask_appbuilder/security/manager.py
fredthomsen: There doesn't seem to be, setting the `ldap.OPT_REFERRALS` prevents referrals from being chased, but not from being returned.  I based my thought process off this [email](https://mail.python.org/pipermail/python-ldap/2014q1/003350.html) in the `python-ldap` maiiling list, which recommended this approach I took above, granted this email is rather old.

fredthomsen: Looks like the python code that wraps the c-extension code in `python-ldap` hides this information away, and there seems to be now way to filter this out in a more explicit manner.  Looking at how the surround code treats a `None` dn and given that we return `(None, None)` when we get an empty search, there is no change in behavior this causes.  We could perhaps split this up and do the filtering on another line if we want to log the original search results.

fredthomsen: Ok, I've made some adjustments to the original search results are still logged and the filtering is more specific.

fredthomsen:
> 
> 
> I also think we should add some unit tests for this, but possibly that is overkill.
> 
> BTW, for unit tests you will need to add refs to [our LDAP test structure](https://github.com/dpgaspar/Flask-AppBuilder/blob/v3.2.2/flask_appbuilder/tests/_test_auth_ldap.py#L68), similar to how [dalibo does it](https://github.com/dalibo/ldap2pg/blob/5.5/tests/unit/test_manager.py#L10-L15), but do note that they use a different LDAP test library.

I tend to agree that unit tests are likely overkill here, and using `mockldap` would be a no go, and I'd have to mock the `search_s` method directly.

thesuperzapper:
@dpgaspar this looks good, can you test this, and then __put out a patch release ASAP__, as this was a breaking change for some users.

The airflow community decided to push the default version of Flask-AppBuilder to `3.2.X`, but this has caused some issues for people who are not setting `AUTH_LDAP_SEARCH`.

/cc @potiuk

dpgaspar (approved):
LGTM

dpgaspar:
Just released 3.2.3rc1

potiuk:
Can we expect an officiall pypi release soon? I am preparing new constraint files for Airflow (to fix other problems) and I would love to include that one

dpgaspar:
yes probably this week

potiuk:
I see. I push the new constraints in ~ 20 minutes. so i will limit it to 3.1.1 for now (but won't add install_requires limits). Thanks for acting quickly !

Issue #1603, opened 2021-04-07
User self-registration fails if belonging to multiple LDAP groups mapping to same role


fredthomsen:
### Environment
Running airflow 2.0.1-python3.8 in docker-compose w/image extended to use flask-appbuilder 3.2.0.
docker-compose version 1.27.4, build unknown
Docker version 19.03.13, build 4484c46d9d
Python 3.8.7

Flask-Appbuilder version: 3.2.0

pip freeze output:
adal==1.2.6
aiohttp==3.7.3
alembic==1.5.4
amqp==2.6.1
apache-airflow==2.0.1
apache-airflow-providers-amazon==1.1.0
apache-airflow-providers-celery==1.0.1
apache-airflow-providers-cncf-kubernetes==1.0.1
apache-airflow-providers-docker==1.0.1
apache-airflow-providers-elasticsearch==1.0.1
apache-airflow-providers-ftp==1.0.1
apache-airflow-providers-google==1.0.0
apache-airflow-providers-grpc==1.0.1
apache-airflow-providers-hashicorp==1.0.1
apache-airflow-providers-http==1.1.0
apache-airflow-providers-imap==1.0.1
apache-airflow-providers-microsoft-azure==1.1.0
apache-airflow-providers-mysql==1.0.1
apache-airflow-providers-postgres==1.0.1
apache-airflow-providers-redis==1.0.1
apache-airflow-providers-sendgrid==1.0.1
apache-airflow-providers-sftp==1.1.0
apache-airflow-providers-slack==2.0.0
apache-airflow-providers-sqlite==1.0.1
apache-airflow-providers-ssh==1.1.0
apispec==3.3.2
appdirs==1.4.4
argcomplete==1.12.2
async-timeout==3.0.1
attrs==20.3.0
azure-batch==10.0.0
azure-common==1.1.26
azure-core==1.11.0
azure-cosmos==3.2.0
azure-datalake-store==0.0.51
azure-identity==1.5.0
azure-keyvault==4.1.0
azure-keyvault-certificates==4.2.1
azure-keyvault-keys==4.3.1
azure-keyvault-secrets==4.2.0
azure-kusto-data==0.0.45
azure-mgmt-containerinstance==1.5.0
azure-mgmt-core==1.2.2
azure-mgmt-datalake-nspkg==3.0.1
azure-mgmt-datalake-store==0.5.0
azure-mgmt-nspkg==3.0.2
azure-mgmt-resource==15.0.0
[...]

PR #1605, opened 2021-04-07
fix: Handle integrity fails if groups map to same roles

fredthomsen:
### Description

If `AUTH_USER_REGISTRATION` is enabled and `AUTH_ROLE_MAPPING` is set such
that multiple LDAP groups map to the same role, then upon attempting to
register the user, specifically inserting that user's roles into the
database, then these inserts will fail due to the contraint check.

To ensure this does not occur, the roles that are to be inserted for the
user needs to be de-dupicated.

Fixes issue #1603.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


fredthomsen:
> 
> 
> @fredthomsen instead of editing the `test__direct_bind__unregistered__multi_role` test, could you make an explicit new test for the constraint failure behaviour? (Which describes what its testing in the docstring & function-name)

Yup.  I've broken this out as a separate test.  This same code could be hit via the oauth path as well, but it's the same underlying path, so I didn't add a test on that side.

flask_appbuilder/tests/_test_auth_ldap.py
```diff
@@ -320,6 +340,30 @@ def test__inactive_user(self):
...
+        sm = self.appbuilder.sm
+
+        # validate - no users are registered
+        self.assertEqual(sm.get_all_users(), [])
+
+        # attempt login
+        user = sm.auth_user_ldap("natalie", "natalie_password")
+
+        # validate - user was allowed to log in
+        self.assertIsInstance(user, sm.user_model)
```
thesuperzapper, line 365: After this can you add the following checks after this:
1. "user was registered" (e.g. [link](https://github.com/dpgaspar/Flask-AppBuilder/blob/v3.3.0/flask_appbuilder/tests/_test_auth_ldap.py#L746-L747))
2. "user was given the correct roles" (e.g. [link](https://github.com/dpgaspar/Flask-AppBuilder/blob/v3.3.0/flask_appbuilder/tests/_test_auth_ldap.py#L749-L753))
3. "user was given the correct attributes (read from LDAP)" (e.g. [link](https://github.com/dpgaspar/Flask-AppBuilder/blob/v3.3.0/flask_appbuilder/tests/_test_auth_ldap.py#L755-L758))
4. "expected LDAP methods were called" (e.g. [link](https://github.com/dpgaspar/Flask-AppBuilder/blob/v3.3.0/flask_appbuilder/tests/_test_auth_ldap.py#L760-L769))

thesuperzapper (changes requested):
@fredthomsen I know its a bit tedious, but can you resolve the testing changes I asked above?

After that, it looks good to merge @dpgaspar

fredthomsen:
> 
> 
> @fredthomsen I know its a bit tedious, but can you resolve the testing changes I asked above?
> 
> After that, it looks good to merge @dpgaspar

No problem.  Issues should be resolved now.

flask_appbuilder/security/manager.py
```diff
@@ -934,15 +934,15 @@ def _ldap_calculate_user_roles(
             # lookup registration role in flask db
             fab_role = self.find_role(registration_role_name)
             if fab_role:
-                user_role_objects.append(fab_role)
+                user_role_objects.add(fab_role)
```
thesuperzapper, line 937: @fredthomsen what happens if the `AUTH_USER_REGISTRATION_ROLE` overlaps with one of the mapped roles?

Does the set() correctly de-dupe the role model objects which have the same name?

fredthomsen: Yes it does.  Since `fab_role` is just the role name the set de-dups them.  We can make this a part of the test case very easily.  I've updated the test case to add the test user to the `AUTH_USER_REGISTRATION_ROLE`.

dpgaspar (approved):
LGTM, @thesuperzapper would be great if you can give it a final pass

thesuperzapper:
@dpgaspar as long as your happy that all possible `role_models` (that you can think of), will be correctly de-duped by `set()`, it looks good to me

As discussed here: https://github.com/dpgaspar/Flask-AppBuilder/pull/1605#discussion_r633147252

fredthomsen:
> 
> 
> @dpgaspar as long as your happy that all possible `role_models` (that you can think of), will be correctly de-duped by `set()`, it looks good to me
> 
> As discussed here: [#1605 (comment)](https://github.com/dpgaspar/Flask-AppBuilder/pull/1605#discussion_r633147252)

@thesuperzapper not sure I understand the concern here.  I thought as long as all roles have unique names, then there shouldn't be a problem here.  Am I missing something?

thesuperzapper:
@fredthomsen I am saying I dont know if all possible `Role` model objects are correctly de-duped by `set()`, that is to say, I haven't checked (it might work fine).

The default Role models are:
* SQLA: https://github.com/dpgaspar/Flask-AppBuilder/blob/v3.3.0/flask_appbuilder/security/sqla/models.py#L58-L68
* Mongo: https://github.com/dpgaspar/Flask-AppBuilder/blob/v3.3.0/flask_appbuilder/security/mongoengine/models.py#L55-L67

fredthomsen:
@thesuperzapper so I am unsure what the underlying `__hash__` and `__eq__` magic methods are used here to ensure that the set members are unique; however, the unit test we have here proves that it works when roles share the same name.

thesuperzapper:
@fredthomsen in my limited testing (not real world, but directly instantiating the `Role` models), the `set()` does not correctly de-dupe:
```
from flask_appbuilder.security.sqla.models import Role as Role

role_1 = Role(id=1, name="test")
role_2 = Role(id=1, name="test")
role_set = set([role_1, role_2])

print(len(role_set))

## output: 2 
## NOTE: this should be 1 if de-dupe works
```

This is why I didn't use `set()` when I first implemented the role mapping feature.

PS: I might be fundamentally misunderstanding something, so If your PR works in the real world (with LDAP setup), please discuss (I am also not sure why it works in the unit-tests, so that might be a place to start looking)

fredthomsen:
@thesuperzapper so contrast what you have with the following snippet which is more in line with how the code and test works:
`
public_1 = sm.find_role("Public")
public_2 = sm.find_role("Public")

assert public_1 == public_2
assert id(public_1) == id(public_2)

print({public_1, public_2})

# output: {Public}
`

Since `BaseSecurityManager.find_role` is used in `BaseSecurityManager.get_roles_from_keys` and in the test itself, that gives more clarity as to why this works.  So this relies on the code for whatever DB abstraction you are using returning objects of the same identity.

thesuperzapper:
Thanks @fredthomsen I am pretty sure you are correct, my test was not representative of how we are retrieving the role objects with `find_role`, then from my end this looks good to go!

thesuperzapper:
@dpgaspar I think this is good to be merged now, its a pretty important fix, and is probably worthy of a patch release.

thesuperzapper:
@fredthomsen I think you need to rebase so @dpgaspar  can merge, but I am not sure?

fredthomsen:
@thesuperzapper @dpgaspar rebased, although I don't think it was needed.

thesuperzapper:
@dpgaspar can you merge?

PR #1608, opened 2021-04-13
fix: relax AzureAD mandatory fields

hyunjong-lee:
### Description

When `given_name` field does not exist in AzureAD, users can't login.
I explicitly claimed `given_name` field in AzureAD, however, `given_name` field does not exist because it is empty.
Therefore, the changes are to relax mandatory fields in AzureAD.

Fix the issue: #1612

> 2021-04-13 04:53:27,972:ERROR:flask_appbuilder.security.views:Error returning OAuth user info: 'given_name'

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -606,7 +606,6 @@ class BaseSecurityManager(AbstractSecurityManager):
                 "last_name": me.get("family_name", ""),
                 "id": me["oid"],
                 "username": me["oid"],
-                "roles": me.get("roles", []),
             }
         # for OpenShift
         if provider == "openshift":
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -603,7 +603,7 @@ class BaseSecurityManager(AbstractSecurityManager):
                 "name": me.get("name", ""),
                 "email": me["upn"],
                 "first_name": me.get("given_name", ""),
-                "last_name": me("family_name", ""),
+                "last_name": me.get("family_name", ""),
                 "id": me["oid"],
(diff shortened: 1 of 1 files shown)


hyunjong-lee:
Currently, I made another pypi project to fix AzureAD auth issue.
- https://pypi.org/project/Flask-AppBuilder-AzureAD/

We are integrating AzureAD auth method in Superset which uses Flask-Appbuilder.
Let me know, if there is something to fix in the code.  @dpgaspar

Thank you! :)

dpgaspar (approved):

flask_appbuilder/security/manager.py
```diff
@@ -600,12 +600,13 @@ def get_oauth_user_info(self, provider, resp):
...
-                "name": me["name"],
+                "name": me.get("name", ""),
                 "email": me["upn"],
-                "first_name": me["given_name"],
-                "last_name": me["family_name"],
+                "first_name": me.get("given_name", ""),
+                "last_name": me.get("family_name", ""),
                 "id": me["oid"],
                 "username": me["oid"],
+                "roles": me.get("roles", []),
```
dpgaspar, line 609: Why the roles here?

dpgaspar (commented):
Left a question, regarding the role field being populated from Azure

flask_appbuilder/security/manager.py
hyunjong-lee: I thought it can be used later like customized security manager, and I found my thought is wrong.
I will delete the line

dpgaspar (approved):

PR #1629, opened 2021-05-05
feat: Add support for before_request hooks

benjreinhart:
### Description

This PR adds support for a `before_request` hook that can be used in FAB's view and API classes to execute some code just before the route handler is invoked. It has the same semantics as Flask's `before_request`, but it is scoped only to the routes within a given FAB class and can be scoped even further to a subset of handlers within the class.

### Examples

```python
@before_request
def ensure_feature_is_enabled(self):
    if self.feature_is_disabled:
        return self.response_404()
    return None
```
```python
@before_request(only=["create", "update", "delete"])
def ensure_write_mode_enabled(self):
    if self.read_only:
        return self.response_400()
    return None
```

### Motivation

_It seems like we could implement the conditional logic directly in route handlers. Why do we need a new pattern?_

In many cases, implementing the conditional logic in the handler itself is preferred. To justify this change, consider the following scenario:

> You've implemented many classes that inherit from `ModelView` or `ModelRestAPI` and have relied on FAB to implement the endpoint logic for you. Some of these classes you only want enabled if some dynamic condition is true at _request time_ (e.g., feature flags, A/B testing). Since you have not implemented the handlers yourself, you have nowhere to put the conditional logic... that is, unless you manually override/implement each one.
>
>In other classes, you have a large number of endpoints that you have implemented. Some need to be gated by a condition. The `before_request` hook provides a way to DRY up your condition handling that is consistent with classes whose handlers are implemented by FAB.

While technically possible to workaround these issues, the hook provides an elegant pattern for developers that enables consistency across classes, a way to keep some endpoint logic DRY and decoupled from handlers, and prevents the need to override/implement FAB's automatic route handling in some cases.

### FAB Menus/Separators

_FAB supports constructing UI menu items and separators, which is registered with its view. If the `before_request` hook  in a view is able to dynamically enabled/disable a route, wouldn't we want the same for the menu item?_

I would imagine that in most (all?) cases, if a route was disabled the menu item should be unavailable too. However, I believe that exposing one pattern that handles both Flask endpoints/responses and UI component availability is hard to do elegantly.

Therefore, I propose treating the request handling and menu item separately. This PR implements the request piece, and a subsequent PR I plan to open will implement an option for executing a callable to determine if menu item is available.

### DIFF
--- a/flask_appbuilder/hooks.py
+++ b/flask_appbuilder/hooks.py
@@ -1,28 +1,40 @@
 def before_request(*args, **kwargs):
     """
-        Use this decorator to add a before_request lifecycle
-        hook to be executed before each handler in the view.
-        If invoked with the `only` kwarg, the hook will only
-        be invoked for the given list of handler methods.
+        Use this decorator to enqueue methods to be invoked
+        before each handler in the view. If method returns
+        a value other than :code:`None`, then that value
+        will be returned to the client. If invoked with the
+        :code:`only` kwarg, the hook will only be invoked for
+        the given list of handler methods.
 
-        Examples:
+        Examples::
 
-            @before_request
-            def ensure_feature_is_enabled(self):
-                if self.feature_is_disabled:
-                    return self.response_404()
-                return None
+            class MyFeature(ModelView)
 
-            @before_request(only=["create", "update", "delete"])
-            def ensure_write_mode_enabled(self):
-                if self.read_only:
-                    return self.response_400()
-                return None
+                @before_request
+                def ensure_feature_is_enabled(self):
+                    if self.feature_is_disabled:
+                        return self.response_404()
+                    return None
 
-        :param only:
-            Only kwarg scopes this hook to run only for
-            the given list of handler methods. If absent,
-            the hook will be invoked before all handlers.
+                # etc...
+
+
+            class MyView(ModelRestAPI):
+
+                @before_request(only=["create", "update", "delete"])
+                def ensure_write_mode_enabled(self):
+                    if self.read_only:
+                        return self.response_400()
+                    return None
+
+                # etc...
+
+
+        :param kwargs:
+            The :code:`only` kwarg scopes the method being decorated
+            to run only for the given list of handler methods. If
+            abset, the method will be invoked before all handlers.
     """
     only = kwargs.get("only", None)
 
--- a/flask_appbuilder/hooks.py
+++ b/flask_appbuilder/hooks.py
@@ -1,11 +1,17 @@
-def before_request(*args, **kwargs):
+from typing import Any, Callable, Dict, List
+
+
+def before_request(
+    hook: Callable[[], Any] = None, only: List[str] = None
+) -> Callable[..., Any]:
     """
-        Use this decorator to enqueue methods to be invoked
-        before each handler in the view. If method returns
-        a value other than :code:`None`, then that value
-        will be returned to the client. If invoked with the
-        :code:`only` kwarg, the hook will only be invoked for
-        the given list of handler methods.
+        This decorator provides a way to hook into the request
+        lifecycle by enqueueing methods to be invoked before
+        each handler in the view. If the method returns a value
+        other than :code:`None`, then that value will be returned
+        to the client. If invoked with the :code:`only` kwarg,
+        the hook will only be invoked for the given list of
+        handler methods.
 
         Examples::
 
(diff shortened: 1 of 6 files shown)


nytai (approved):
LGTM, aside from a couple minor lint comments

flask_appbuilder/hooks.py
```diff
@@ -0,0 +1,76 @@
+def before_request(*args, **kwargs):
```
dpgaspar, line 1: FAB is not totally type annotated, but every new addition should be.

flask_appbuilder/hooks.py
```diff
@@ -0,0 +1,76 @@
...
+    return wrapped_handler
+
+
+def get_before_request_hooks(view_or_api_instance):
+    before_request_hooks = []
+    for attr_name in dir(view_or_api_instance):
+        attr = getattr(view_or_api_instance, attr_name)
+        if hasattr(attr, "_before_request_hook") and attr._before_request_hook is True:
+            before_request_hooks.append(attr)
+    return before_request_hooks
```
dpgaspar, line 76: totally optional, a list comprehension could look good here and can be slightly more efficient.

flask_appbuilder/hooks.py
```diff
@@ -0,0 +1,76 @@
...
+
+                # etc...
+
+
+        :param kwargs:
+            The :code:`only` kwarg scopes the method being decorated
+            to run only for the given list of handler methods. If
+            abset, the method will be invoked before all handlers.
+    """
+    only = kwargs.get("only", None)
```
dpgaspar, line 39: Why not make the signature like `before_request(only: Optional[List[str]]) -> Callable[..., Any]:`

dpgaspar (approved):
Looks really good!

- Let's also add tests for `ModelView` classes with `before_request`
- Add a reference on the docs for the `ModelView` also

This supports multiple `before_request` is there a use case for it? Should we add tests with multiple `before_request` ?

dpgaspar (commented):
.

dpgaspar (commented):
seems like tests began to fail. let me take a look at that

benjreinhart (commented):
I don't see any failing tests, where do you see them failing?

flask_appbuilder/hooks.py
benjreinhart: 👍🏻

flask_appbuilder/hooks.py
benjreinhart: It's not as simple as that since it can be used as a decorator without being invoked by user (in that case, `args[0]` is a bound function) or it can be invoked by user with `only` kwarg which returns a function that takes a bound function.

I'll see what I can do for this.

dpgaspar:
> I don't see any failing tests, where do you see them failing?

No worries, was a temporary failure on packages microsoft

benjreinhart:
@dpgaspar I added more comprehensive tests and updated to add type hints. As for the docs, I am happy to make more updates, but where I added it is the API reference section, which AFAICT is general API reference, not tied to FAB API vs model. Let me know what you think.

flask_appbuilder/hooks.py
benjreinhart: 👍🏻

dpgaspar:
@benjreinhart update to `37`
https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/tests/test_mvc.py#L586

flask_appbuilder/tests/test_api.py
```diff
@@ -2788,3 +2837,137 @@ class Model1PermConverge(ModelRestApi):
...
+        """
+        client = self.app.test_client()
+        token = self.login(client, USERNAME_ADMIN, PASSWORD_ADMIN)
+
+        # Everything is disabled
+        self._before_request_condition_value = False
+        self._before_request_can_read = False
+        self._before_request_can_write = False
+
+        uri = "api/v1/model1beforerequest/basic"
```
dpgaspar, line 2853: optional: we could do a for loop here to DRY these tests a bit

benjreinhart: Personally, I like keeping test code "damp" rather than DRY because I think that writing out each case explicitly is easier to read, serving as documentation. Also, less likely to introduce a bug into tests if you avoid "programming" in tests.

That being said, I am happy to change it if you feel strongly.

PR #1631, opened 2021-05-07
feat: Support for conditional menu item rendering

benjreinhart:
### Description

This is "part 2" of https://github.com/dpgaspar/Flask-AppBuilder/pull/1629. It adds support for conditionally rendering menu items at request time.

@dpgaspar the tests for the menu piece specifically are not very comprehensive but I think we should get more confidence around this code. Do you have a preference for how you'd like to expand upon these tests? The menu code is almost entirely self-contained, so IMO we should pass the one non-self-contained piece (`current_app`) which let's us easily stub that value and unit test the code. That should give us all we need, since this is already tested from API PoV. Thoughts?

Also, let me know what other (if any) docs you think should be added here.

### DIFF
--- a/flask_appbuilder/base.py
+++ b/flask_appbuilder/base.py
@@ -356,6 +356,7 @@ class AppBuilder(object):
         category="",
         category_icon="",
         category_label="",
+        menu_cond=None,
     ):
         """
         Add your views associated with menus using this method.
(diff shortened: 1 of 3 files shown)


craig-rueda (approved):

flask_appbuilder/base.py
```diff
@@ -464,7 +481,12 @@ def add_link(
             :param category_label:
                 The label that will be displayed on the menu,
                 if absent param name will be used
-
+            :param cond:
+                If a callable, :code:`cond` will be invoked when
+                constructing the menu items. If it returns :code:`True`,
+                then this link will part of the menu. Otherwise, it
```
dpgaspar, line 487: `then this link will be part of the menu`

dpgaspar (approved):
LGTM, the API get's self documented

Issue #1627, opened 2021-04-29
Dissuade users from modifying roles through UI if AUTH_ROLES_SYNC_AT_LOGIN is enabled


cccs-rc:
### Environment

Flask-Appbuilder version: 3.2.1

<details>
  <summary>pip freeze output</summary>
aiohttp==3.7.2
alembic==1.4.3
amqp==2.6.1
apispec==3.3.2
async-timeout==3.0.1
attrs==20.2.0
Authlib==0.15.2
Babel==2.8.0
backoff==1.10.0
billiard==3.6.3.0
bleach==3.2.1
boto3==1.16.10
botocore==1.19.10
Brotli==1.0.9
cached-property==1.5.2
cachelib==0.1.1
celery==4.4.7
certifi==2020.6.20
cffi==1.14.3
chardet==3.0.4
click==7.1.2
colorama==0.4.4
contextlib2==0.6.0.post1
convertdate==2.3.0
cron-descriptor==1.2.24
croniter==0.3.36
cryptography==3.2.1
decorator==4.4.2
defusedxml==0.6.0
Deprecated==1.2.11
dnspython==2.0.0
elasticsearch==7.12.1
elasticsearch-dbapi==0.2.1
email-validator==1.1.1
et-xmlfile==1.0.1
Flask==1.1.2
Flask-AppBuilder==3.2.1
Flask-Babel==1.0.0
Flask-Caching==1.9.0
Flask-Compress==1.8.0
Flask-Cors==3.0.9
Flask-JWT-Extended==3.24.1
Flask-Login==0.4.1
Flask-Migrate==2.5.3
Flask-OpenID==1.2.5
Flask-SQLAlchemy==2.4.4
flask-talisman==0.7.0
Flask-WTF==0.14.3
future==0.18.2
geographiclib==1.50
geopy==2.0.0
gunicorn==20.0.4
holidays==0.10.3
humanize==3.1.0
idna==2.10
ijson==3.1.2.post0
importlib-metadata==2.1.1
isodate==0.6.0
itsdangerous==1.1.0
jdcal==1.4.1
Jinja2==2.11.3
jmespath==0.10.0
jsonlines==1.2.0
jsonschema==3.2.0
kombu==4.6.11
korean-lunar-calendar==0.2.1
linear-tsv==1.1.0
Mako==1.1.3
Markdown==3.3.3
MarkupSafe==1.1.1
marshmallow==3.9.0
marshmallow-enum==1.5.1
[...]

PR #1642, opened 2021-05-18
fix: add warning text to roles when AUTH_ROLES_SYNC_AT_LOGIN

dpgaspar:
### Description

Adds a warning message when creating or editing a user and `AUTH_ROLES_SYNC_AT_LOGIN` is enabled.

Screenshot 2021-05-20 at 09 51 29

cc: @cccs-rc @thesuperzapper 

### ADDITIONAL INFORMATION
- [X] Has associated issue: #1627
- [X] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -124,8 +124,10 @@ class UserInfoEditView(SimpleFormView):
 def _roles_custom_formatter(string: str) -> str:
     if current_app.config.get("AUTH_ROLES_SYNC_AT_LOGIN", False):
         string += (
-            ". <b>Warning: AUTH_ROLES_SYNC_AT_LOGIN is enabled. "
-            "Changes to this field will not persist."
+            ". <div class='alert alert-warning' role='alert'>"
+            "AUTH_ROLES_SYNC_AT_LOGIN is enabled, changes to this field will "
+            "not persist between user logins."
+            "</div>"
         )
     return string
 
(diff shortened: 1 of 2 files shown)


flask_appbuilder/security/views.py
```diff
@@ -119,6 +121,15 @@ def form_post(self, form):
         flash(as_unicode(self.message), "info")
 
 
+def _roles_custom_formatter(string: str) -> str:
+    if current_app.config.get("AUTH_ROLES_SYNC_AT_LOGIN", False):
+        string += (
+            ". <b>Warning: AUTH_ROLES_SYNC_AT_LOGIN is enabled. "
+            "Changes to this field will not persiste."
```
cccs-rc, line 128:
```suggestion
            "Changes to this field will not persist."
```
Just a typo! :) So happy to see this getting added!

dpgaspar: nice catch, thks!

cccs-rc (approved):

flask_appbuilder/security/views.py
```diff
@@ -119,6 +121,15 @@ def form_post(self, form):
         flash(as_unicode(self.message), "info")
 
 
+def _roles_custom_formatter(string: str) -> str:
+    if current_app.config.get("AUTH_ROLES_SYNC_AT_LOGIN", False):
+        string += (
+            ". <b>Warning: AUTH_ROLES_SYNC_AT_LOGIN is enabled. "
+            "Changes to this field will not persist."
```
thesuperzapper, line 128: I think "Changes to this field will not persist between user logins." is more correct

thesuperzapper (changes requested):
@dpgaspar should we make the colour more obvious, say red instead of black?

PR #1660, opened 2021-06-23
fix: Following "next" URL after login.

suiting-young:
### Description

The `has_access` decorator provides a "next" parameter when redirecting to this endpoint.
https://github.com/dpgaspar/Flask-AppBuilder/blob/b9b3259a9d47a14430144c8362215c2143f06aa4/flask_appbuilder/security/decorators.py#L117-L122
Should try to redirect back here if login succeeded.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -532,7 +532,8 @@ class AuthLDAPView(AuthView):
                 flash(as_unicode(self.invalid_login_message), "warning")
                 return redirect(self.appbuilder.get_url_for_login)
             login_user(user, remember=False)
-            return redirect(self.appbuilder.get_url_for_index)
+            next_url = request.args.get("next")
+            return redirect(next_url or self.appbuilder.get_url_for_index)
         return self.render_template(
             self.login_template, title=self.title, form=form, appbuilder=self.appbuilder
         )


dpgaspar (commented):
looks good, can you add some tests

suiting-young:
I have added a parent class `AuthBasicView` for both `AuthDBView` and `AuthLDAPView`.
As the `login` method of them are extremely same, except the auth method.
And furthermore, I don't know how to test `AuthLDAPView` 😝 . Codecov tells me that my test is against `AuthDBView`.

---
_P.S._ I removed `from urllib(.parse) import quote` Python2 fallback from `test_mvc.py`, because other 5 files only use Python3 module structure.

suiting-young:
@dpgaspar could you please review again, thanks.

suiting-young:
Hi @dpgaspar,

I have rebased the PR to latest `master`.
And there is only one thing left that need to be fixed then:

If the user failed to login,
- it should stay at the **current** page (_**login** actually_),
  with any parameter (_e.g. `?next=...`_) retained;
- instead of redirecting to purified `get_url_for_login`.

Please have a look.
Thanks

PR #1662, opened 2021-06-25
feat: add import/export of roles with permissions

krsnik93:
### Description

This PR introduces two new CLI commands:

- `flask fab export-roles`
- `flask fab import-roles`

In combination, they allow for quick migration of custom roles with permissions from one host to another.

I initially tried using `Model2SchemaConverter` to create schemas for `sqla` models. After making a couple of changes, I was able to use it to serialize nested models. However, I was not able to de-serialize back from this result, so I resorted to creating individual `marshmallow` schemas.

I don't think `export_roles` and `import_roles` should live in `SecurityManager` classes, but not sure where to put them. Would appreciate feedback.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [x] Introduces new feature


dpgaspar (commented):
Looking good, but missing:
- tests, unit-tests and cli
- User's can define builtin roles these should be included also

krsnik93:
@dpgaspar regarding builtin roles, I can see that they get inserted into the database when the `SecurityManager` is initialized:
https://github.com/dpgaspar/Flask-AppBuilder/blob/b9b3259a9d47a14430144c8362215c2143f06aa4/flask_appbuilder/security/manager.py#L778-L779
So as long as the user initializes the app or runs it, the builtin roles will be in the database as well and will get exported. I tried adding an entry for `FAB_ROLES` in `config.py` and running `flask run` followed by `flask fab export-roles` and the resulting file had the builtin roles included as well.

flask_appbuilder/tests/test_fab_cli.py
```diff
@@ -70,3 +75,46 @@ def test_list_views(self):
...
+            dst_app.config["FAB_ROLES"] = {}
+            dst_db = SQLA(dst_app)
+            dst_app_builder = AppBuilder(dst_app, dst_db.session)
+            dst_cli_runner = dst_app.test_cli_runner()
+
+            filename = "roles.json"
+
+            export_result = src_cli_runner.invoke(export_roles, [f"--path={filename}"])
+            self.assertEqual(export_result.exit_code, 0)
+            self.assertTrue(os.path.exists(filename))
```
dpgaspar, line 104: Would be great if we could assert the content of the file also.
nit: can you split these into 2 distinct tests? by assert the content you can add the file has a fixture

flask_appbuilder/tests/test_fab_cli.py
```diff
@@ -70,3 +75,46 @@ def test_list_views(self):
...
+
+            import_result = dst_cli_runner.invoke(import_roles, [f"--path={filename}"])
+            self.assertEqual(import_result.exit_code, 0)
+
+            # after import roles on dst app include all the roles from src app,
+            # including the ones from FAB_ROLES in config
+            self.assertEqual(
+                len(dst_app_builder.sm.get_all_roles()),
+                len(src_app_builder.sm.get_all_roles()),
+            )
```
dpgaspar, line 120: can you also assert the permissions on the DB after the import?

dpgaspar (commented):
looking good, left a couple of comments regarding tests

krsnik93:
@dpgaspar I think there might be some issues with the code coverage tool as it flags lines that are certainly covered...

dpgaspar (approved):
LGTM!
Would be great if you could drop a few lines here: https://flask-appbuilder.readthedocs.io/en/latest/cli.html

krsnik93:
@dpgaspar do you agree about the coverage being off or should I take a look?

dpgaspar:
@krsnik93 something is strange, I'm not finding your tests on: https://pipelines.actions.githubusercontent.com/EcVmXJ9QlFNYqdgzb19Iib0zlAKatWUkmR9lWHntlezNeYEMGz/_apis/pipelines/1/runs/716/signedlogcontent/13?urlExpires=2021-07-06T12%3A29%3A02.1318847Z&urlSigningMethod=HMACV1&urlSignature=FzwDirCfFm%2FHEn%2Bzk4HTwZu%2FStetzk99ZRo5YR4xl3o%3D

Can you take a look, ping me if you can't figure what's wrong

krsnik93:
Hi @dpgaspar the link does not work for me, says "Uri expired"

krsnik93:
@dpgaspar you are right that `nose` isn't picking up all the tests, in fact it's not picking up `test_fab_cli.py` at all. I am working on a fix.

krsnik93:
@dpgaspar I managed to make `nose` discover all tests. I have everything passing locally, but `mongodb` fails here. I added a new container into `docker-compose` which is needed for the new tests, but I am not sure it's used properly. Could you please try it out locally?

krsnik93:
@dpgaspar looking good now

dpgaspar:
@krsnik93 great work!

PR #1663, opened 2021-06-27
fix: check if there is an email field in userinfo

chakku000:
### Description
When using OAuth authorization provider which doesn't give email information, `oauth_authorized` method cause KeyError exception. This is because at [Line714](https://github.com/dpgaspar/Flask-AppBuilder/blob/b9b3259a9d47a14430144c8362215c2143f06aa4/flask_appbuilder/security/views.py#L714), userinfo's `email` key is accessed even if userinfo does not have `email` key.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar (approved):
Thank you for the fix

PR #1668, opened 2021-07-12
fix: improve next URL on OAuth

dpgaspar:
### Description

Improves logic on next URL OAuth. Adds type annotations

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -712,7 +712,6 @@ class AuthOAuthView(AuthView):
                 parsed_uri = urlparse(state["next"][0])
                 if parsed_uri.netloc != request.host:
                     log.warning("Got an invalid next URL: %s", parsed_uri.netloc)
-                    next_url = self.appbuilder.get_url_for_index
                 else:
                     next_url = state["next"][0]
             return redirect(next_url)
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -1,8 +1,20 @@
 import datetime
 import logging
 import re
+from typing import Optional
+from urllib.parse import urlparse
 
-from flask import abort, current_app, flash, g, redirect, request, session, url_for
+from flask import (
+    abort,
+    current_app,
+    flash,
+    g,
+    redirect,
+    request,
+    Response,
+    session,
+    url_for,
+)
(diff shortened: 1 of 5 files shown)


dpgaspar:
@ashb if you have time a review would be nice

flask_appbuilder/security/views.py
```diff
@@ -735,11 +706,15 @@ def oauth_authorized(self, provider):
...
-            try:
-                next_url = state["next"][0] or self.appbuilder.get_url_for_index
-            except (KeyError, IndexError):
-                next_url = self.appbuilder.get_url_for_index
-
+            next_url = self.appbuilder.get_url_for_index
+            # Check if there is a next url on state
+            if "next" in state and len(state["next"]) > 0:
+                parsed_uri = urlparse(state["next"][0])
+                if parsed_uri.netloc != request.host:
```
ashb, line 713: Is it possible to check the "base url" instead? I can't remember if that information makes it down to FAB or not (I suspect not?) but if it does checking that would be better.

Also should we check port number too? (which I guess could be optional/empty on either side?)

flask_appbuilder/security/views.py
```diff
@@ -735,11 +706,15 @@ def oauth_authorized(self, provider):
...
+                    log.warning("Got an invalid next URL: %s", parsed_uri.netloc)
+                    next_url = self.appbuilder.get_url_for_index
```
ashb, line 715: Already set outside the check.

```suggestion
```

flask_appbuilder/security/views.py
ashb: Oh `request.host` already includes non-standard ports

dpgaspar: yes, so I think it's fine, do you agree?

ashb: Yeah request.host is good.


Is it worth checking `parsed_uri.path` starts with https://flask.palletsprojects.com/en/2.0.x/api/?highlight=request#flask.Request.root_path for  too?

ashb (approved):
This is a good improvement, I'll let you decide if it's also worth checking `root_path`

PR #1679, opened 2021-08-20
other: Redirect to the referrer URL when changing locale.

tianhe1986:
Signed-off-by: tianhe1986 <w1s2j3229@163.com>

### Description

In most case, users hope just refreshing the current page after changing locale.
However, not all pages call `update_redirect`, turning out it will jump to other page at the top of the page history stack.
For example, [this issue in apache superset](https://github.com/apache/superset/issues/12768)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


flask_appbuilder/babel/views.py
```diff
@@ -13,6 +14,12 @@ class LocaleView(BaseView):
...
             abort(404, description="Locale not supported.")
+
+        if request.referrer is not None:
+            page_history = Stack(session.get("page_history", []))
+            page_history.push(request.referrer)
+            session["page_history"] = page_history.to_json()
+
         session["locale"] = locale
         refresh()
         self.update_redirect()
```
dpgaspar, line 25: The `LocaleView.index` is already calling `update_redirect` why updating it again? because we need to update it with the referrer?

Also would it make more sense to add the `update_redirect` to Superset?

dpgaspar (commented):
Thank you for the PR, left a couple of questions

flask_appbuilder/babel/views.py
tianhe1986: > The LocaleView.index is already calling update_redirect why updating it again? because we need to update it with the referrer?

Yes,  updating it with the referrer before calling `update_redirect` would ensure it redirects back to the referrer page.

> Also would it make more sense to add the update_redirect to Superset?

Surely you are right:) It is just a bit tough to add `update_redirect` in every related view, and the same case even with 
 build-in pages, not only Superset.

For example, with the URL "/login/", that is, `login` function at subclass of `class AuthView`, e.g [class AuthDBView](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/views.py#L513),  it also does not call the `update_redirect` and would redirect to other page while changing locale.

junlincc:
Thank you for the contribution! Would love to get it through. @villebro want to chime in?

Issue #1576, opened 2021-02-26
[feature request] Password policy/Complexity


kamalkeshavani-aiinside:
Currently Flask AppBuilder does not allow the password check for any complexity requirements, so we rely on the end users to setup a strong password.

It would be great if we could force some minimum complexity requirements in the password, so that end users can not set weak password.

dpgaspar:
Agree this should be included. 

I'll make a config hook, with a simple default, that people can modify. Maybe a hookable field validator

kamalkeshavani-aiinside:
@dpgaspar Can you please tag this to avoid getting stale label?

dpgaspar:
@kamalkeshavani-aiinside new PR about to be merged will support custom complexity and default complexity. Feel free to comment on the PR

dpgaspar:
@kamalkeshavani-aiinside thank you for your help, the PR is merged and will be released soon

PR #1687, opened 2021-09-06
feat: password complexity option on DB Auth

dpgaspar:
### Description

New feature to support password complexity validation to AUTH database users.

Adds the following new optional config keys:

- FAB_PASSWORD_COMPLEXITY_VALIDATOR: Hook for a custom function to validate password complexity
- FAB_PASSWORD_COMPLEXITY_ENABLED: Enables new FAB default password complexity validator

This PR also adds partial mypy type checking and a small test refactor

### ADDITIONAL INFORMATION
- [X] Has associated issue: #1576 
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar:
@ashb @potiuk

docs/security.rst
```diff
@@ -602,6 +602,38 @@ exclude them from add and edit form. Using our example you will define our view
...
+This feature only makes sense when using AUTH database.
+By default you can enable password complexity validation by setting `FAB_PASSWORD_COMPLEXITY_ENABLED = True`.
+
+This default enforces:
+
+- At least 2 Uppercase letters
+- At least 3 Lowercase letters
+- At least 1 special character
+- At least 2 numeric digits
+- At least 10 total digits
```
kamalkeshavani-aiinside, line 617: It should be 'At least 10 total characters'.

dpgaspar: fixed, thks!

kamalkeshavani-aiinside: LGTM, it has all the options I was looking for.

flask_appbuilder/security/forms.py
```diff
@@ -5,6 +5,7 @@
 
 from ..fieldwidgets import BS3PasswordFieldWidget, BS3TextFieldWidget
 from ..forms import DynamicForm
+from ..validators import PasswordComplexityValidator
```
villebro, line 8: Just an opinion, but while we're on the topic of improving typing/readability, I somehow always find fully qualified imports more pleasant to read:

```python
from flask_appbuilder.validators import PasswordComplexityValidator
```
over
```python
from ..validators import PasswordComplexityValidator
```

flask_appbuilder/validators.py
```diff
@@ -1,26 +1,46 @@
...
+from flask_appbuilder.exceptions import PasswordComplexityValidationError
+from flask_appbuilder.models.base import BaseInterface
+from flask_babel import gettext
+from wtforms import Field, Form, ValidationError
 
-class Unique(object):
+password_complexity_regex = re.compile(
+    r"""(
+    ^(?=.*[A-Z].*[A-Z])                # at least two capital letters
+    (?=.*[!@#$&*])                     # at least one of these special characters
```
villebro, line 13: To not "force" users to use a highly restricted set of special characters, would it be better to do a negative match for non-special characters here? Something like 
```regex
(?!([a-zA-Z0-9])).
```

villebro (commented):
One personal opinion + one minor improvement proposal

flask_appbuilder/validators.py
dpgaspar: good point

flask_appbuilder/security/forms.py
dpgaspar: Currently I prefer them also, but I'm following the current code base "convention", I plan to change them on a different PR

villebro (approved):
LGTM

potiuk:
Oh cool. did not have time to follow up on this one. let me take a look.

flask_appbuilder/models/base.py
```diff
@@ -47,6 +47,13 @@ class BaseInterface(object):
     def __init__(self, obj: Type[Any]):
         self.obj = obj
 
+    def __getattr__(self, name: str) -> Any:
```
potiuk, line 50: Nice. I did not know about it !

flask_appbuilder/security/forms.py
potiuk: Agree. Relative imports are evil.

potiuk: Coming as strong advise in Airflow 2.2: https://github.com/apache/airflow/blob/main/docs/apache-airflow/modules_management.rst#dont-use-relative-imports

flask_appbuilder/tests/security/test_password_complexity.py
```diff
@@ -0,0 +1,32 @@
+from flask_appbuilder.exceptions import PasswordComplexityValidationError
+from flask_appbuilder.validators import default_password_complexity
+
+from ..base import FABTestCase
+
+
+class PasswordComplexityTestCase(FABTestCase):
```
potiuk, line 7: Suggestion: Maybe you could use https://pypi.org/project/parameterized/ here?

potiuk: And few other places.

flask_appbuilder/validators.py
```diff
@@ -31,3 +51,41 @@ def __call__(self, form, field):
...
+    def __call__(self, form: Form, field: Field) -> None:
+        password_complexity_validator = current_app.config.get(
+            "FAB_PASSWORD_COMPLEXITY_VALIDATOR", None
+        )
+        if password_complexity_validator is not None:
+            try:
+                password_complexity_validator(field.data)
+            except PasswordComplexityValidationError as exc:
+                raise ValidationError(str(exc))
+        if current_app.config.get("FAB_PASSWORD_COMPLEXITY_ENABLED", False):
```
potiuk, line 71: I think the current set of configuration variables is a bit confusing.

When I see `FAB_PASSWORD_COMPLEXITY_ENABLED` - think this parameter configures password complexity check in general, so I expect that if `FAB_PASSWORD_COMPLEXITY_ENABLED` is "false" then no validation is done (even if I have my own custom validator configured in `FAB_PASSWORD_COMPLEXITY_VALIDATOR`).

Currently, in order to enable ONLY custom validator you have to have effectively this combination:

```
FAB_PASSWORD_COMPLEXITY_ENABLED = false
FAB_PASSWORD_COMPLEXITY_VALIDATOR = your_validator
```

This is pretty confusing (because I'd expect `FAB_PASSWORD_COMPLEXITY_ENABLED = false` disables validation of password completely.

I think two one of the two approaches would be better:

a) `FAB_PASSWORD_COMPLEXITY_ENABLED` - enables validation in general, and `FAB_PASSWORD_COMPLEXITY_VALIDATOR` replaces the default validator with the custom one. In this case you cannot run both custom and default validation at the same time

b) `FAB_PASSWORD_COMPLEXITY_ENABLED` - should be renamed to `FAB_PASSWORD_DEFAULT_COMPLEXITY_VALIDATOR` - to not make impression that whole feature is disabled by setting it to `false`

I have slight preference for option a) (I think it's better to always run one validator - either default or custom, you can always derive your custom validator from the default one and use it internally) - but I think both a) and b) are better than current proposal.

potiuk:
Thanks for that @dpgaspar - I had some other pressing issues in the meantime, so I did not pick it up as promised, sorry for that. I hoe you find my comments useful

flask_appbuilder/tests/security/test_password_complexity.py
dpgaspar: did not know that one, only this: https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions.

dpgaspar: added looks better and it encourages to write more test cases (which I did) :)

flask_appbuilder/validators.py
dpgaspar: Very good point, went for option a)

dpgaspar:
> Thanks for that @dpgaspar - I had some other pressing issues in the meantime, so I did not pick it up as promised, sorry for that. I hoe you find my comments useful

No worries at all, I had this one on my TODO list also. Thank you for the review and yes the comment were great, looks much better now!

flask_appbuilder/validators.py
potiuk: Cool. Looks great now. Then I have one more NIT comment :)

docs/config.rst
```diff
@@ -285,6 +285,13 @@ Use config.py to configure the following parameters. By default it will use SQLL
 | FAB_STATIC_URL_PATH                    | Path to override default static folder     |           |
 |                                        |                                            |   No      |
 +----------------------------------------+--------------------------------------------+-----------+
+| FAB_PASSWORD_COMPLEXITY_VALIDATOR      | Hook for your own custom password          |           |
+|                                        | validator function.                        |   No      |
++----------------------------------------+--------------------------------------------+-----------+
+| FAB_PASSWORD_COMPLEXITY_ENABLED        | Enables the default password complexity    |           |
```
potiuk, line 291:
```suggestion
| FAB_PASSWORD_COMPLEXITY_ENABLED        | Enables the password complexity            |           |
```

potiuk (commented):
Small NIT after choosing option a)

potiuk (approved):

potiuk:
Fantastic!

potiuk:
🎉 🎉 🎉 🎉 🎉

PR #1698, opened 2021-09-14
fix: DateTimePicker rendering in forms

### LINKED ISSUE #685: DateTime field shows a "DatePicker" only?
I've seen the fix in #22 and it seems to work as I can see the correct css class etc. when I add a DateTime field to my model (using sqlalchemy and sqlite3):
`<div class="input-group date appbuilder_datetime" id="datetimepicker">`

Nevertheless I can not choose a time (hour/minute) via the widget - only a date.
If I click on a date the current time together with the selected date gets inserted into the field. Is this normal?

[screenshot](https://ibb.co/f3Mbew)

fedepad:
### Description

`DateTimePicker` renders fine (date plus time) for filters but not for forms (e.g. add) with `DateTime` fields.
This seems to be mainly due to disabling the ability to pick times (`{pickTime: false}`) in `ab.js` but not in `ab_filter.js`
This fix uniforms `ab.js` datetimepicker options to the ones (working) in `ab_filter.js`
I don't know if there are any gotchas or other things to change to make it work, but to me it looks this is the first thing to change to grant the ability for the picker to not disable picking times.
Also it wasn't clear how to test this since this would be kind of like UI test but any suggestions are appreciated.

This fix refers to the closed but still unresolved issue:
https://github.com/dpgaspar/Flask-AppBuilder/issues/685

### ADDITIONAL INFORMATION

- [x] Has associated issue: Fixes #685


fedepad:
I just found this: https://github.com/dpgaspar/Flask-AppBuilder/pull/871
What was the problem back then?

fedepad:
On a form I am displaying, using the latest flask appbuilder with this patch, it shows the following images for the datetime field/picker rendering and can change also the time.
Attached screenshots.
Screenshot 2021-09-15 at 11 38 48
Screenshot 2021-09-15 at 11 39 04
Screenshot 2021-09-15 at 11 40 59

aquaino:
I noticed the same problem. Why this pull request have not yet been accepted?

aquaino (approved):

dpgaspar (approved):
Tested and looks great, Thank you

allilou:
I have this same issue with the latest version 4.3.11
It's caused by a change in `ab.js` in the following commit

happycode-2010:
Oh，I used FAB version is 4.5.1, this problem is  reappears,  i solved by  back the version to 4.3.7, same, allilou.

PR #1706, opened 2021-09-30
fix oauth when looking up by email

itajaja:
should fix https://github.com/dpgaspar/Flask-AppBuilder/commit/dbe1eded6369c199b777836eb08d829ba37634d7#r57324441

### Description

I encountered this regression that probably I only have :) but still seems like a bug was introduced. this happens when users are matched by email and not by username.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


itajaja:
nevermind, I see how this is problematic down the line. I fixed on my end by updating the usernames to be the email addresses :v:

PR #1707, opened 2021-10-04
fix: OAuth login flow

dpgaspar:
### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -604,15 +604,12 @@ class AuthOAuthView(AuthView):
             return redirect(self.appbuilder.get_url_for_index)
 
         if provider is None:
-            if len(self.appbuilder.sm.oauth_providers) > 1:
-                return self.render_template(
-                    self.login_template,
-                    providers=self.appbuilder.sm.oauth_providers,
-                    title=self.title,
-                    appbuilder=self.appbuilder,
-                )
-            else:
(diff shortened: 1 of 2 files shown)


shawnzhu:
@dpgaspar I like this change because:

1. it improves the sign-in flow for users who use OAuth2 provider(s), just like any other websites supports social login buttons.
2. make the oauth2 provider name explicit
3. could log out users (by redirecting back to the sign in page).

Removal of the <kbd>Register</kbd> button makes sense. Since people will take it as alternative of OAuth2 provider which is not the case at all when using `AUTH_TYPE = OAUTH2_AUTH`

shawnzhu:
The only comments from me is this is a break change of sign in UX (no more auto sign-in, new <kbd>Sign in with xxx :partying_face: </kbd> button) which needs to be documented, so I assume there will be documentation change as well (happy to help if needed!)

shawnzhu:
Just tried this feature from v3.4.0, works great and thank you @dpgaspar !

dpgaspar:
that's great, thank you for ALL your help, really appreciated

mgorsk1:
I just dug out this PR as this change has influenced our user flow. I wonder whether this is something community would consider bringing back but maybe as a opt-in behavior?

PR #1730, opened 2021-11-02
feat: Add groups to azure oauth2 provider role_keys

andersbjht:
### Description

Add mapping from oauth groups to role_keys for azure.

It was only implemented for certain providers, like okta.

### ADDITIONAL INFORMATION

In order to export "groups", also need to add "openid".

The default in Azure is to use the Group IDs as keys.


dpgaspar (approved):
LGTM, thank you!

hterik:
@dpgaspar is there anything remaining before this can get merged?

georgewfisher (changes requested):

PR #1737, opened 2021-11-09
feat: allow multiple values for the same filter

willrogers:
### Description

Allow multiple filters to apply to the same filter-column combination - for example `will` in `email_address` and `gmail` in `email_address`.

This is an update of #791, which was a fix for #789 but was not finally completed. I have added some tests as requested and will happily update docs with a little guidance.

### ADDITIONAL INFORMATION

- [x] Has associated issue:
- [ ] Is CRUD MVC related.
- [x] Introduces new feature


willrogers:
@dpgaspar this CI failure is unexpected since I have not changed dependencies. It appears that something has changed meaning that setuptools 58 or greater is being used, and that is not compatible with funcparserlib: https://github.com/vlasovskikh/funcparserlib/issues/65

There is a `1.0.0a1` release that would fix this, but until then it is possible for CI to run successfully if you use an older version of setuptools: https://github.com/willrogers/Flask-AppBuilder/commit/7c28b3384d1fc29c4542148f0b3a9df628f9a535

If I'm correct (I could be wrong) you will start seeing this on all pull requests.

ThomasP0815:
Filter value is concenated using comas if i use the same operator twice for a certain field

issue_1737
issue_1737_2

Searching again delivers:
issue_1737_3

willrogers:
@ThomasP0815 thanks, I will have a look next week.

willrogers:
@ThomasP0815 thanks for reviewing, I believe that problem is resolved. I am generally happy to take recommendations on the best way to do things though.

ThomasP0815:
@willrogers nice job, now it works perfect. thanks a lot.

ThomasP0815 (approved):

runoutnow:
Tried this out, but it looks like some classes in models/sqla/filters.py does need an update too...

error log:
flask_appbuilder/models/sqla/filters.py", line 125, in apply
    return query.filter(field.ilike("%" + value + "%"))
TypeError: can only concatenate str (not "list") to str

willrogers:
Thanks for testing @runoutnow, I will see if I can reproduce this.

Can you give a bit more detail about what you did to trigger this error?

runoutnow:
> Thanks for testing @runoutnow, I will see if I can reproduce this.
> 
> Can you give a bit more detail about what you did to trigger this error?

Never mind, it was my fault. I only changed the filters.py in my local file, not in the venv... 😬 it works now, although I am unable to use the search function for searching "relation as many" items. The result is always wrong... but that has nothing to do with this... still got to find out if that's a bug or not. 😊

willrogers:
@dpgaspar is this something that you would consider merging? I am happy to make further changes if required.

examples/issue_789/Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
```
dpgaspar, line 1: examples should run with the current project requirements file. Can you please delete these?

willrogers: Sure, I will do.

willrogers: I have done that and update the README as well.

dpgaspar (commented):
Looks good, missing tests

willrogers:
> Looks good, missing tests

@dpgaspar thank you. There are a couple of tests added in `flask_appbuilder/tests/test_mvc.py`, is there something else specific that you would like testing?

dpgaspar (approved):
Nice work!

willrogers:
@dpgaspar is there any prospect of getting this branch merged? It will fix a couple of bugs for us.

dpgaspar:
@willrogers just merged master and tests are failling now. Can you check what could be wrong? I'm planning on merging this and make a new release next week

willrogers:
Thank you, I have fixed the test.

PR #1758, opened 2021-12-13
fix: Set role and confirm password while adding user mandatory

mayurnewase:
### Description

Role and Confirm password field in user db model view were not marked as required field, so added those validators.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar:
Hi, thank you for contributing.

Please install the `requirements-dev.txt` and run `black` to pass CI

dpgaspar (commented):
CI is failling

dpgaspar (commented):
Looks good can you add a test to assert user's can't add a user without a role

dpgaspar (approved):

PR #1766, opened 2022-01-05
fix: Handle authorize_access_token exception

mjkonarski-b:
### Description

Currently the call to `authorize_access_token` in `oauth_authorized` is not wrapped in exception handling. As a result when an error occurs in the OAuth flow, a 500 response is generated and the user sees an error page. This may lead to a bad user experience, because the user is not able to fix the problem in any way on their own.

This PR adds an error handling, so that any error will be caught and the user is redirected to the login page with the appropriate flash error message.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar (approved):

dpgaspar:
makes sense, thank you @mjkonarski-b

PR #1775, opened 2022-01-07
fix: Only update user.last_login on successful authentication

blag:
### Description

This PR moves one line around so the `user.last_login` field is only updated when the user has successfully authenticated.

Without this PR the `user.last_login` field is not very useful, as an attacker trying to brute force a login would continuously update the `last_login` field to the datetime of the latest _unsuccessful_ authentication attempt, instead of the datetime of the last successful authentication attempt. I believe that "login" usually refers to a successful authentication attempt, so this keeps the behavior of the code consistent with the semantics of the field name.

I don't have a thorough understanding of what all `user.last_login` could be/is used for, but I can imagine that, when it records the last datetime of successful authentication attempt, it can be combined with `user.fail_login_count` to generate an average "brute force rate" for each user between successful logins. Without this PR, it is impossible to collect that data because the `user.last_login` will almost always be set to the datetime of the latest brute force attempt.

I have expanded the docstring to have a more thorough description of the method, and I have added tests for each line and branch of the `BaseSecurityManager.update_user_auth_stat` method so it is now completely covered by tests.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -820,12 +820,17 @@ class BaseSecurityManager(AbstractSecurityManager):
 
     def update_user_auth_stat(self, user, success=True):
         """
-        Update authentication successful to user.
+        Update user authentication stats upon successful/unsuccessful
+        authentication attempts.
 
         :param user:
-            The authenticated user model
+            The identified (but possibly not successfully authenticated) user
+            model
         :param success:
-            Default to true, if false increments fail_login_count on user model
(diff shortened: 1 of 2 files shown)


dpgaspar (approved):

dpgaspar:
Agree, thank you @blag

kaxil:
Can we merge this one @dpgaspar :)

PR #1783, opened 2022-01-12
feat: SQLAlchemy 1.4 support

### LINKED ISSUE #1710: Relax SQLAlchemy version dependency
Right now in `setup.py` we have SQLAlchemy pinned to <1.4.0:

        # SQLAlchemy 1.4.0 breaks flask-sqlalchemy and sqlalchemy-utils
        "SQLAlchemy<1.4.0",

flask-sqlalchemy added support for 1.4 in release [2.5.0](https://github.com/pallets/flask-sqlalchemy/blob/main/CHANGES.rst#version-250) (released in March).
sqlalchemy-utils added support for 1.4 in release [0.37.0](https://github.com/kvesteri/sqlalchemy-utils/blob/master/CHANGES.rst#0370-2021-04-12) (released in April).

I think we should update SQLAlchemy dependency.

uranusjr:
### Description

An alternative way to fix #1710. Flask-SQLAlchemy 2 broke the `clause=None` parameter, but since `mapper=None, clause=None` is simply the default since SQLAlchemy 0.7.2 (released more than ten years ago), we can simply fix the problem by not passing those parameters at all.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #1710
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/cli.py
+++ b/flask_appbuilder/cli.py
@@ -142,7 +142,7 @@ def create_db():
     """
     from flask_appbuilder.models.sqla import Model
 
-    engine = current_app.appbuilder.get_session.get_bind(mapper=None, clause=None)
+    engine = current_app.appbuilder.get_session.get_bind()
     Model.metadata.create_all(engine)
     click.echo(click.style("DB objects created", fg="green"))
 
--- a/flask_appbuilder/console.py
+++ b/flask_appbuilder/console.py
@@ -197,7 +197,7 @@ def create_db(app, appbuilder):
     from flask_appbuilder.models.sqla import Base
 
     _appbuilder = import_application(app, appbuilder)
-    engine = _appbuilder.get_session.get_bind(mapper=None, clause=None)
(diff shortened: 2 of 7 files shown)


dpgaspar (commented):
Hi, this would be great, but while testing it I got:
```
DB Creation and initialization failed: get_bind() got an unexpected keyword argument 'bind'
```
I think that get_bind from flask-sqlalchemy is still incompatible with SQLAlchemy 4.X

uranusjr:
I regenerated the entire `requirements.txt` with pip-compile instead. I think this should work?

uranusjr:
Oh, since all test jobs share the same requirements.txt, the 3.6 test does not run anymore since some of the dependencies have dropped 3.6 (since it already reached EOL). What is a good approach to this? IMO the test jobs should _not_ share requirements.txt because the file depends on Python version. They should instead directly install the package from `setup.py`. If a fully-pinned requirements file is desired, each Python version should independently generate one instead.

PR #1787, opened 2022-01-14
feat: Add gte and lte filters for sqla

fredthomsen:
### Description

Add <= and >= filters for SQLA.  These seem common enough that including them here makes sense.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/models/sqla/filters.py
+++ b/flask_appbuilder/models/sqla/filters.py
@@ -183,6 +183,34 @@ class FilterSmaller(BaseFilter):
         return query.filter(field < value)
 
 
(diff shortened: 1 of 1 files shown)


fedepad:
The tests seem to fail because you need to update the field_* here: https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/tests/test_api.py#L1783

by adding the new filters for the field_* types for which you defined them in the conversion table.
As far as I can see from your code you will need to add for field_date, field_integer, field_float.
If you need to update other tests you will see.

PR #1724, opened 2021-10-25
feat(cli): allow `export-roles` to be beautified

EBoisseauSierra:
### Description

Expose `json.dump(indent=*)` to the `fab export-roles --indent=*` CLI command.

### ADDITIONAL INFORMATION

- [x] Has associated issue: #1716 
- [ ] Is CRUD MVC related.
- [ ] Is Auth, RBAC security related.
- [ ] Changes the security db schema.
- [ ] Introduces new feature
- [ ] Removes existing feature

PR #1811, opened 2022-02-21
test: Add test for `export-roles --indent`'s argument “duck casting” to int

EBoisseauSierra:
### Description

#1724 introduced the `--indent` argument to `fab export-roles`. This (optional) argument accepts `int` and `str` args — and notably *negative* integers.

However, `fab export-roles --indent=-1` would naturally return `-1`… as a `str`, and not a negative `int`. That's why we [“duck cast” the arg](https://github.com/EBoisseauSierra/Flask-AppBuilder/blob/ba7da50cb40aa41c07b7642494ff18ce0e390eb6/flask_appbuilder/cli.py#L159-L166) to an integer… if possible.

This piece of logic remained however untested. So for the sake of the exercise, this PR:

* extracts the duck casting logic into an ad hoc method,
* create tests for this method.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/cli.py
+++ b/flask_appbuilder/cli.py
@@ -28,6 +28,21 @@ def echo_header(title):
     click.echo(click.style("-" * len(title), fg="green"))
 
 
+def cast_int_like_to_int(cli_arg: Union[None, str, int]) -> Union[None, str, int]:
+    """Cast int-like objects to int if possible
+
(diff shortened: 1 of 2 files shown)


dpgaspar (approved):
Awesome! thank you

PR #1812, opened 2022-02-23
feat: An optional route to allow authentication and registration via api requests

Mimameid:
### Description

Added an optional `AuthApi` which allows authentication interaction via rest endpoints.

Currently, there is no inbuilt option to use the session based authentication functionality without enabling the `ModelView` based [security models](https://github.com/dpgaspar/Flask-AppBuilder/tree/master/flask_appbuilder/security).This comes along with a strong coupling to the UI rendering provided by FAB.

Using the `AuthApi` the client can now leverage session tokens set by the server with the `httpOnly` flag by issuing http requests and register new users, everything decoupled from the rendering.
 
This is a first step to allow a client (or another webserver) handle the rendering.
Doesn't implement email verification.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [x] Introduces new feature


dpgaspar (commented):
Can you please add more details about this on the description. Also missing tests

Thank you!

Mimameid:
> Can you please add more details about this on the description. Also missing tests
> 
> Thank you!

Hi Daniel,

updated the description. I hope it is more clear now, what this is about. If not, please get back to me (if possible with specific questions). Regarding tests, you are right. I will add them.

PR #1846, opened 2022-05-10
fix: Set certificates before reconnecting to LDAP

sbernauer:
### Description
Hi Flask-AppBuilder team,

we set up Superset - using Flask-AppBuilder - to use an LDAP connection as described in [your documentation](https://flask-appbuilder.readthedocs.io/en/latest/security.html#authentication-ldap).
The first time a user tries to log into the Superset instance the login will always fail with the following error shown in the logs:
```
2022-05-10 13:45:43,341:ERROR:flask_appbuilder.security.manager:{'result': -1, 'desc': "Can't contact LDAP server", 'ctrls': [], 'info': 'error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed (self signed certificate in certificate chain)'}
```
The second login would than pass (if the credentials are correct).
This patches solves the issue, so that the first login attempt passes.

[The docs of python-ldap](https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html) say
> libldap does not materialize all TLS settings immediately. You must use OPT_X_TLS_NEWCTX with value 0 to instruct libldap to apply pending TLS settings and create a new internal TLS context:

We have to set all the certificates before we set OPT_X_TLS_NEWCTX to 0 so that the correct certificates are set.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar (approved):
thank you!

PR #1897, opened 2022-07-27
fix: fix a wrong 'next' URL in javascript

sansarun:
### Description

Due to Jinja autoescaping, when we render the `next` url in html the `&` become escaped to `&amp;`
This result in a wrong URL parameters.

For example
`?next=https://www.abc.com?id=1&name=a` will be rendered as `?next=https://www.abc.com?id=1&amp;name=a` in HTML page.

The fix is to hint Jinja not to escape the URL and also encode the URL.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar:
thank you @sansarun!

Issue #804, opened 2018-09-05
Unable to "filter" or implement equivalent of SQLite "WHERE" claude on Views


chatlourd:
I would like to be able to implement the equivalent of SQLite "WHERE" clause on a `ModelView`:

`SELECT Field_A
FROM Table_A
WHERE Field_A = 'some text'`

I have read a lot about queries and filters but have so far been un able to apply them to `Flask-AppBuilder`. Let's assume we have the most simple Model:

`class Table_A(Model):
    id = Column(Integer, primary_key=True)
    Field_A =  Column(String)

    def __repr__(self):
        return self`

and View:

`class Table_AView(ModelView):
    datamodel = SQLAInterface(Table_A)
    label_columns = {'Field_A':'A'}
    list_columns = ['Field_A']`

How would I be able to apply a `WHERE` clause so that my ModelView restrict my table to lines where Field_A is equal to "some text" ?

Issue #803, opened 2018-08-31
Add terms for registration


francisliyy:
![image](https://user-images.githubusercontent.com/6333979/44915586-59533d00-ad01-11e8-83bb-4761d51c5166.png)

Is there a way to add terms before reCaptcha for registration page?

dolfandringa:
Sorry @francisliyy for the late reply, but it is possible. You should implement your own Security Manager (see https://flask-appbuilder.readthedocs.io/en/latest/security.html#your-custom-security), and override the registerusedbview (or registeruseroidview depending on you authentication method) to point to your own view. Then implement your own view, that inherits from the relevant flask_appbuilder.security.registerviews.RegisterUser(DB/OID)View and changes the form attribute to point to your own form, that again, inherits from flask_appbuilder.security.forms.RegisterUser(DB/OID)Form. That is just a regular Flask Appbuilder form, so you can  override it's add_template (see https://flask-appbuilder.readthedocs.io/en/latest/templates.html#add-templates) to include anything you like.

dolfandringa:
If you're willing to document this properly, a pull request to [docs/security.rst](https://github.com/dpgaspar/Flask-AppBuilder/blob/develop/docs/security.rst) would be very welcome.

PR #1903, opened 2022-08-11
chore: Increase upper-bound on apispec

tm-drtina:
### Description
My local tests (postgres and mongodb) pass with apispec 5.2.2 (currently latest).
I don't think we need to limit `apispec<4`. If there are issues for consumers of this lib, they can put stronger constraints to their requirements.
Still kept some upper bound to avoid any breaking change from future major versions.

Also had to upgrade `pip-tools`, since current version breaks on pip progress bar

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar (commented):
Thank you for the PR.

Are you using `ModelRestApi` on your tests? have you checked the OpenAPI spec?

tm-drtina:
I've ran `nosetests -v flask_appbuilder.tests.test_api` which should use the `ModelRestApi` and all tests passed.

I haven't checked the OpenAPI spec.

I don't think newer apispec breaks anything in how FAB uses it. It still might break something in consumers of this lib, but they can simply put more strict requirements in their setup.cfg

tm-drtina:
According to the changelog following are backwards incompatible changes:
- Change MarshmallowPlugin.map_to_openapi_type from a decorator to a classic function, taking a field as first argument ([#804](https://github.com/marshmallow-code/apispec/pull/804)).
- Remove validate_spec from public API. Users may call their validator of choice directly ([#803](https://github.com/marshmallow-code/apispec/pull/803)).
- Remove OpenAPIVersion and use packaging.Version instead ([#801](https://github.com/marshmallow-code/apispec/pull/801)).

I don't see any of these used in the codebase, so I think it shouldn't break the lib, but still can break consumers.
I still think it should be ok to increase the version range and let consumers limit the version range themselves.

dpgaspar (approved):

Issue #1908, opened 2022-08-17
Password should not return in to_json


luutruong:
There is very critical when use method `to_json` in model User; it also includes `password` into response.

youngsol:
@dpgaspar  answered in here
https://github.com/dpgaspar/Flask-AppBuilder/pull/1909#pullrequestreview-1096148147

"use marshmallow to serialize data"

PR #1909, opened 2022-08-17
fix: Support remove some fields on method to_json

luutruong:
### Description

This MR, I implemented new logic which it allows model can define hidden fields when cast to json.

For example:

```python

class ExampleModel(Model):
  id = Column(...)
  secret_column = Column(...)

  json_hidden_fields = ['secret_column']


example = ExampleModel()
json = example.to_json() # should not include the column secret_column

```

### ADDITIONAL INFORMATION

- [ ] Has associated issue: https://github.com/dpgaspar/Flask-AppBuilder/issues/1908
- [ ] Is CRUD MVC related.
- [x] Introduces new feature

### DIFF
--- a/flask_appbuilder/models/sqla/__init__.py
+++ b/flask_appbuilder/models/sqla/__init__.py
@@ -121,10 +121,18 @@ class Model(object):
     """
 
     __table_args__ = {"extend_existing": True}
+    """
+        Remove fields when invoke to_json method.
+        For example: json_hidden_fields = ['name', 'foo', 'baz']
+    """
+    json_hidden_fields = None
 
     def to_json(self):
         result = dict()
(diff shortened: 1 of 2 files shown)


dpgaspar (commented):
When a user defines a `Model` with a field named `json_hidden_fields` it will probably break
Also `to_json` is a utility currently used on FAB's MVC, would using marshmallow for serialisation work for you?

luutruong:
@dpgaspar You're right with conflict with column named `json_hidden_fields` but we can change this field to another `__json_hidden_fields__` (as example).

> Also to_json is a utility currently used on FAB's MVC, would using marshmallow for serialisation work for you?

I think this is simple filter so it does not need other dependencies.

PR #1922, opened 2022-09-08
fix: add_view menu_cond

steve-embling:
### Description

menu_cond parameter of appbuilder.add_view is described as being a callable or None, that will hide the menu item if returns false. But setting this as per docs doesn't hide the menu item.

Apologies for lack of test case. It looks like current ones fail if an item is artificially hidden, but doesn't fail if an item is not hidden.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/filters.py
+++ b/flask_appbuilder/filters.py
@@ -1,5 +1,5 @@
 from flask import current_app, request, url_for
-
+import sys
 from .const import PERMISSION_PREFIX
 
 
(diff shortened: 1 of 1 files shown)


steve-embling:
Closing this - It doesnt only hide from the main menu but from action menus too, so I don't think that was the intention. Will step back and take a look for a more sensible approach, probably in menu

steve-embling:
Rereading the docs and actually struggling to interpret exactly what is intended, so maybe this does meet it.

It's not what I was looking for (keeping the views but leaving them out of the main menu) but appears to meet conditions described in docs, kind of! Sorry.

I do not think this should be merged but I'll reopen for discussion perhaps.

PR #1926, opened 2022-09-15
fix: Check if Oauth login with OKTA is correct

safeith:
### Description

We should check the errors for OKTA Oauth; if we do not check it, it will continue and will create a user with the default information that we create in the return clause.
image

Why does this happen?
Sometimes users are in more than 100 groups and Okta does not return all the groups and it return the following error message
```
User info from Okta: {'error': 'server_error', 'error_description': 'The groups claim matched too many groups and must be configured to match fewer groups.'}
```

It returns an error but, we do not check it and we let it to continue with the following return clause
```
return {
    "username": "okta_",
    "first_name": "",
    "last_name": "",
    "email": "",
    "role_keys": [],
}
```

And with data with register a new user(s)  [here](https://github.com/dpgaspar/Flask-AppBuilder/blob/master/flask_appbuilder/security/manager.py#L1272)

Finally we have imagery user that many users can login with it

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -660,13 +660,17 @@ class BaseSecurityManager(AbstractSecurityManager):
             me = self.appbuilder.sm.oauth_remotes[provider].get("userinfo")
             data = me.json()
             log.debug("User info from Okta: %s", data)
-            return {
-                "username": f"{provider}_{data['sub']}",
-                "first_name": data.get("given_name", ""),
-                "last_name": data.get("family_name", ""),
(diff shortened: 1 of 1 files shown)


dpgaspar (approved):

dpgaspar (commented):
Sorry for the delay here, could you please rebase and add a couple of tests. Thank you!

safeith:
Hi @dpgaspar 
No worries, I rebased it.

PR #1936, opened 2022-10-12
fix: Save next URL on failed login attempt

Dosenpfand:
### Description

The login process for DB and LDAP does not save the next URL parameter after one failed login attempt.
When trying to access protected view a redirect to `/login` with the get parameter `next` set to the URL of the protected view happens. When the user tries with a wrong username and password combination the first time, the user gets redirected again to `/login` but without the `next` parameter. So if the user successfully logs in after the 1st attempt, no redirect to the protected view happens.

Steps to reproduce:
1. Go to http://flaskappbuilder.pythonanywhere.com/users/list/
2. Try to log in with incorrect user:password combination: guest:not_welcome
3. Log in with correct combination: guest:welcome
4. No redirect to http://flaskappbuilder.pythonanywhere.com/users/list/ happens

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar (commented):
thank you! can you add tests?

dpgaspar (approved):
Sorry for the delay...

Thank you once more, highly appreciated

PR #1956, opened 2022-12-08
fix: Replace deprecated attachment_filename

### LINKED ISSUE #1947: Download issue in view.py
[download in view.py](https://github.com/dpgaspar/Flask-AppBuilder/blob/d6f06fbd6532ff86bd215a52bfc37f04a867603d/flask_appbuilder/views.py#L643) is using send_file from flask.

Here is send_file definition:

```python
def send_file(
    path_or_file: t.Union[os.PathLike, str, t.BinaryIO],
    mimetype: t.Optional[str] = None,
    as_attachment: bool = False,
    download_name: t.Optional[str] = None,
    conditional: bool = True,
    etag: t.Union[bool, str] = True,
    last_modified: t.Optional[t.Union[datetime, int, float]] = None,
    max_age: t.Optional[
        t.Union[int, t.Callable[[t.Optional[str]], t.Optional[int]]]
    ] = None,
) -> "Response":
```

And is passing `attachment_filename` instead of `download_name`

I did the patch and it works!

steve-embling:
### Description

Replace attachment_filename with download_name, as it has been removed from flask. Fixes #1947 

### ADDITIONAL INFORMATION

- [X] Has associated issue:
- [X] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/views.py
+++ b/flask_appbuilder/views.py
@@ -640,7 +640,7 @@ class ModelView(RestCRUDView):
     def download(self, filename):
         return send_file(
             op.join(self.appbuilder.app.config["UPLOAD_FOLDER"], filename),
-            attachment_filename=uuid_originalname(filename),
+            download_name=uuid_originalname(filename),
             as_attachment=True,
         )
 


dpgaspar (approved):
Great, thank you!

PR #1976, opened 2023-01-09
feat: Add rate limiter

bolkedebruin:
### Description

This allows regulation of the number of times a user can request a particular API or view given a time-frame. It protects against brute-forcing entry and denial of service.

Both Airflow and Superset have a long standing issue of not being able to rate limit access to APIs and views allowing for brute forcing access.

https://github.com/apache/superset/issues/10560
https://lists.apache.org/thread/hdpyr3y5cq0djcogn1ql7gj7vq09xo0j
https://github.com/apache/superset/issues/10620

This change allows users to decorate their APIs and Views with `@limit`.

Note I have turned to default to be ON for authentication for AuthDB and AuthLDAP (Oauth/oidc have their own protection mechanisms) as FAB should be secure out of the box I assume.

cc @potiuk

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [X] Introduces new feature


flask_appbuilder/security/manager.py
```diff
@@ -285,6 +291,14 @@ def __init__(self, appbuilder):
         # Setup Flask-Jwt-Extended
         self.jwt_manager = self.create_jwt_manager(app)
 
+        # Setup Flask-Limiter
+        self.limiter = self.create_limiter(app)
+
+    def create_limiter(self, app) -> Limiter:
```
dpgaspar, line 297: nit: `def create_limiter(self, app: Flask) -> Limiter:`

flask_appbuilder/security/manager.py
```diff
@@ -489,6 +503,14 @@ def openid_providers(self):
     def oauth_providers(self):
         return self.appbuilder.get_app.config["OAUTH_PROVIDERS"]
 
+    @property
+    def is_auth_limited(self):
```
dpgaspar, line 507: nit: `def is_auth_limited(self) -> bool:`

flask_appbuilder/security/manager.py
```diff
@@ -489,6 +503,14 @@ def openid_providers(self):
...
+        return self.appbuilder.get_app.config["AUTH_RATE_LIMITED"]
+
+    @property
+    def auth_rate_limit(self):
```
dpgaspar, line 511: nit: `def auth_rate_limit(self) -> str:`

flask_appbuilder/security/decorators.py
```diff
@@ -228,3 +231,68 @@ def wraps(f):
...
+     deduction should be done from the rate limit
+    :param on_breach: a function that will be called when this limit
+     is breached. If the function returns an instance of :class:`flask.Response`
+     that will be the response embedded into the :exc:`RateLimitExceeded` exception
+     raised.
+    :param cost: The cost of a hit or a function that
+     takes no parameters and returns the cost as an integer (Default: ``1``).
+    """
+
+    def wraps(f):
```
dpgaspar, line 282: nit: missing type annotations here

flask_appbuilder/security/manager.py
```diff
@@ -1548,6 +1577,22 @@ def get_user_menu_access(self, menu_names: List[str] = None) -> Set[str]:
                 None, "menu_access", view_menus_name=menu_names
             )
 
+    def add_limit_view(self, baseview):
+        if baseview.limits:
```
dpgaspar, line 1581: nit: can you invert this condition the reduce nesting:
``` python
if not baseview.limits:
    return
for limit in baseview.limits:
    self.limiter.limit(
        limit_value=limit.limit_value,
        key_func=limit.key_func,
        per_method=limit.per_method,
        methods=limit.methods,
        error_message=limit.error_message,
        exempt_when=limit.exempt_when,
        override_defaults=limit.override_defaults,
        deduct_when=limit.deduct_when,
        on_breach=limit.on_breach,
        cost=limit.cost,
        )(baseview.blueprint)
```

dpgaspar (commented):
This is great! left some non blocking comments. Can you please revisit?

bolkedebruin:
@dpgaspar ptl, i have addressed your comments and fixed the linting (can I suggest using a newer flake8, the one used here is really really old).

bolkedebruin:
Hi @dpgaspar PR is finished. Would be great if it was included in 4.2.1 :-)

dpgaspar:
sorry, it won't, but I'll make a 4.3.0 next week just for this, no worries

bolkedebruin:
awesome, that'll work :-)

flask_appbuilder/security/manager.py
```diff
@@ -255,6 +257,10 @@ def __init__(self, appbuilder):
             app.config.setdefault("AUTH_LDAP_LASTNAME_FIELD", "sn")
             app.config.setdefault("AUTH_LDAP_EMAIL_FIELD", "mail")
 
+        # Rate limiting
+        app.config.setdefault("AUTH_RATE_LIMITED", True)
+        app.config.setdefault("AUTH_RATE_LIMIT", "2 per 5 second")
```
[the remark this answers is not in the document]
bolkedebruin: I don't think it's aggressive. A human being will have trouble meeting this if they have to retype the password and otherwise a short pause won't hurt. An automated system configured wrongly will remain wrong.

So if you realize "omg ive typed in the password/username twice" you probably start thinking a bit longer and cross the 5s barrier (try it!) and if you do make it a slight pause won't hurt IMHO.

An alternative is to redirect to a captcha when reaching the limit but I'll leave that to someone else.

Anyways I'm not married to this limit but I do not think it's too aggressive.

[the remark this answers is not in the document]
dpgaspar: I was able to reach it pretty easily and manually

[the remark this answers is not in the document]
bolkedebruin: Easily as in normal human behavior or as in trying to trigger it? Triple entry isn't common. 

But if you think it's better we can switch.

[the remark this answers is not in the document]
dpgaspar: Trying to trigger it, but If you agree, I think it would be more conservative and still safe against brute force and DoS

dpgaspar (approved):

PR #1978, opened 2023-01-09
fix: Make sure user input is not treated as safe in the oauth view

gschuurman:
### Description

Fix a security issue issue with the next parameter in the OAUTH Login field.

The user can edit the next field from within the URL, this can cause reflected XSS. 
for example /login/?next="-alert(document.location)-" will alert the document location.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


dpgaspar (approved):

raboof:
Was an advisory ever published for this issue? I don't think I see it at https://github.com/dpgaspar/Flask-AppBuilder/security

PR #2008, opened 2023-03-20
fix:  Actions from different views on same page

Dosenpfand:
### Description

When having a page with different views e.g. a main and one or multiple related views with actions, they do not work as intended as all HTML `<form>`s will have the same ID `action_form`. This PR fixes the issue by appending the model view name to the ID.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/static/appbuilder/js/ab_actions.js
+++ b/flask_appbuilder/static/appbuilder/js/ab_actions.js
@@ -6,7 +6,7 @@
 //------------------------------------------------------
 // AdminActions holds methods to handle UI for actions
 //------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Dosenpfand:
Hi @dpgaspar, is there a chance to get this merged soonish?

Dosenpfand:
@dpgaspar , I applied the fix on the current tip of `master`. Please let me know if there is anything else I can do to get this merged. Thanks!

DBREngineer:
@dpgaspar , It would be very nice to apply this patch.

Thanks

PR #2015, opened 2023-04-13
added comma - from apispec[yaml]>=1.1.1<2 to apispec[yaml]>=1.1.1,<2…

sandeepvarmakanumuri:
… in setup.py file

### Description

setup.py file have a missing comma in the install_requires var with 'apispec[yaml]>=1.1.1<2',. This line has been modified to 'apispec[yaml]>=1.1.1,<2'. Airflow older version puckle/docker-airflow 10.6 have a dependency from flask-appbuilder which installs as a subprocess from 1.13.1 version. So, this is required to be fixed as the dependency is failing the setup.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     install_requires=[
         'colorama>=0.3.9,<1',
         'click>=6.7,<8',
-        'apispec[yaml]>=1.1.1<2',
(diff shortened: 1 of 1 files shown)


dpgaspar:
Good catch, thank you!

PR #2053, opened 2023-06-05
fix: Double escaping for next param in login with oauth

AVMusorin:
### Description

Since next parameter shouldn't be used with `safe` Jinja filter (see https://github.com/dpgaspar/Flask-AppBuilder/pull/1978), to avoid double escaping we can use Jinja's `urlencode`. Also, it might be clearer to work with this parameter in the related libraries, no need to double unquote the url.

cc @gschuurman 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


AVMusorin:
Hi @dpgaspar! Could you please take a look on this PR?

dpgaspar (approved):
LGTM, thank you!

PR #2102, opened 2023-08-14
fix: increase email field length

dpgaspar:
### Description

Email max length should be 320 = 64 + 1 + 255

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/mongoengine/models.py
+++ b/flask_appbuilder/security/mongoengine/models.py
@@ -77,7 +77,7 @@ class User(Document):
     username = StringField(max_length=64, required=True, unique=True)
     password = StringField(max_length=256)
     active = BooleanField()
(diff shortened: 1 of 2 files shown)


john-bodley:
@dpgaspar could you explain the meaning behind the `64 + 1 + 255` calculation? Using Aurora MySQL 5.7 using the InnoDB storage engine we're running into the following key—``UNIQUE KEY `email` (`email`)``—length error, 

```
mysql> ALTER TABLE ab_user MODIFY email VARCHAR(320) NOT NULL;
ERROR 1071 (42000): Specified key was too long; max key length is 767 bytes
mysql> 
```

PR #2112, opened 2023-09-01
fix: Validate Azure JWTs using `authlib`

wolfdn:
### Description

This PR introduces [authlib](https://github.com/lepture/authlib) to decode and validate JWTs from Azure.
Note: I tried to use `joserfc` instead of authlib - [as recommended in the authlib docs](https://docs.authlib.org/en/latest/jose/index.html#jose-guide) - but then realized that joserfc is only compatible with Python >= 3.8. To keep compatibility with Python 3.7 we should use authlib.

Until now we were using a custom implementation to parse Azure JWTs. Only the content of JWTs was evaluated, but it was not checked if the signature of the JWT is even valid.

By using `authlib` we can get rid of the custom parsing implementation for Azure JWTs in `manager.py` and it also allows us to validate the claims of the JWT.
The public keys from Microsoft are automatically fetched and used to decode the JWT.
Some basic claims of the JWT are also validated (`issued at`, `not before` and `expiry`).

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -5,6 +5,7 @@ import logging
 import re
 from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
+import requests
 from flask import Flask, g, session, url_for
 from flask_babel import lazy_gettext as _
 from flask_jwt_extended import current_user as current_user_jwt
@@ -12,6 +13,8 @@ from flask_jwt_extended import JWTManager
 from flask_limiter import Limiter
 from flask_limiter.util import get_remote_address
 from flask_login import current_user, LoginManager
+from joserfc import jwt
+from joserfc.jwk import KeySet
 from werkzeug.security import check_password_hash, generate_password_hash
 
 from .api import SecurityApi
@@ -632,11 +635,9 @@ class BaseSecurityManager(AbstractSecurityManager):
         # https://docs.microsoft.com/en-us/azure/active-directory/develop/
         # active-directory-protocols-oauth-code
         if provider == "azure":
-            log.debug("Azure response received : %s", resp)
-            id_token = resp["id_token"]
-            log.debug(str(id_token))
-            me = self._azure_jwt_token_parse(id_token)
-            log.debug("Parse JWT token : %s", me)
+            log.debug("Azure response received:\n%s", json.dumps(resp, indent=4))
+            me = self._decode_and_validate_azure_jwt(resp["id_token"])
+            log.debug("Decoded JWT:\n%s", json.dumps(me, indent=4))
             return {
                 "name": me.get("name", ""),
                 "email": me["upn"],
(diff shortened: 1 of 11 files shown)


potiuk (approved):

flask_appbuilder/security/manager.py
```diff
@@ -683,36 +682,17 @@ def get_oauth_user_info(self, provider, resp):
...
-
-        if not decoded_payload:
-            log.error("Payload of id_token could not be base64 url decoded.")
-            return
-
-        jwt_decoded_payload = json.loads(decoded_payload.decode("utf-8"))
+    def _decode_and_validate_azure_jwt(self, id_token):
+        keyset = JsonWebKey.import_key_set(
+            requests.get(
+                "https://login.microsoftonline.com/common/discovery/keys"
```
dpgaspar, line 688: what about making this a configuration key or a constant?

dpgaspar (commented):
Looks good and simple, can you add tests please?

setup.py
```diff
@@ -46,6 +46,7 @@ def desc():
     platforms="any",
     install_requires=[
         "apispec[yaml]>=6.0.0, <7",
+        "authlib>=1.2.1",
```
dpgaspar, line 49: Authlib needs to be an extra requirement

flask_appbuilder/security/manager.py
wolfdn: Just moved this URL to the constants file. Since this URL is static for Microsoft OAuth (no matter which tenant is used) I thought it is probably not necessary to make this configurable.

flask_appbuilder/security/manager.py
```diff
@@ -683,36 +682,17 @@ def get_oauth_user_info(self, provider, resp):
...
+            ).json()
+        )
+        claims = jwt.decode(id_token, keyset)
+        claims.validate()
+        log.info("Decoded JWT:\n%s", json.dumps(claims, indent=4))
```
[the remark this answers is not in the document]
wolfdn: Just changed it to `log.debug`

setup.py
wolfdn: I just removed this from `install_requires` - I saw that authlib is already included in the `oauth` section of `extras_require`.

wolfdn:
> Looks good and simple, can you add tests please?

I just added a unittest for the new `_decode_and_validate_azure_jwt` method.
Since this method is relying on `authlib` for JWT decoding and validation, this test is just making sure that the `validate` function of the JWT claims is actually called.

Is this sufficient as a unit test?

requirements.txt
```diff
@@ -6,6 +6,8 @@
 #
 apispec[yaml]==6.3.0
     # via Flask-AppBuilder (setup.py)
+authlib==1.2.1
```
dpgaspar, line 9: can you move this to requirements extra?

dpgaspar (approved):
Looking good one last comment

dpgaspar:
> > Looks good and simple, can you add tests please?
> 
> I just added a unittest for the new `_decode_and_validate_azure_jwt` method. Since this method is relying on `authlib` for JWT decoding and validation, this test is just making sure that the `validate` function of the JWT claims is actually called.
> 
> Is this sufficient as a unit test?

It's ok, I'm going to revisit the Azure and claims mapping also, I'll add more tests

requirements.txt
wolfdn: I noticed that `authlib` is already in `requirements-extra.txt` - now I just upgraded the version there from `0.15.4` to the latest version `1.2.1`

flask_appbuilder/security/manager.py
```diff
@@ -683,36 +683,13 @@ def get_oauth_user_info(self, provider, resp):
...
-        payload_b64_string += "=" * (4 - ((len(jwt_payload) % 4)))
-        decoded_payload = base64.urlsafe_b64decode(payload_b64_string.encode("ascii"))
-
-        if not decoded_payload:
-            log.error("Payload of id_token could not be base64 url decoded.")
-            return
-
-        jwt_decoded_payload = json.loads(decoded_payload.decode("utf-8"))
+    def _decode_and_validate_azure_jwt(self, id_token):
+        keyset = JsonWebKey.import_key_set(requests.get(MICROSOFT_KEY_SET_URL).json())
```
dpgaspar, line 687: this should be cached or set at boot time

dpgaspar:
Thank you @wolfdn !

potiuk:
Cool!

PR #2120, opened 2023-09-30
fix: Remove erroring endpoint "/login/<provider>/<register>" from AuthOAut…

dkalamar:
…hView

### Description
Keyword argument `register` was removed from `AuthOAuthView.login` as part of #1707. However, the endpoint `/login/<provider>/<register>/` is still exposed, resulting in a 500 error whenever it is called:
```
TypeError: AuthOAuthView.login() got an unexpected keyword argument 'register'
```
This PR fixes this error and any calls to that endpoint will properly return a 404 instead.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/views.py
+++ b/flask_appbuilder/security/views.py
@@ -604,7 +604,6 @@ class AuthOAuthView(AuthView):
 
     @expose("/login/")
     @expose("/login/<provider>")
-    @expose("/login/<provider>/<register>")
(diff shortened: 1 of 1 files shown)


dpgaspar (approved):
Good catch

Issue #2137, opened 2023-10-14
Just released WTForms 3.1.0 break FAB widgets (including just released 4.3.8)


potiuk:
The WTForms 3.1.0 introduced a change that breaks implementation of Flask App Builder choices fields based on QuerySelectField.

The problem is that the new choices widget returns now possible "render_kwargs" parameter in "iter_choices" : 

See the release notes for 3.1.0

* https://wtforms.readthedocs.io/en/3.1.x/changes/#version-3-1-0 

The issue in WTForms is: https://github.com/wtforms/wtforms/issues/692
The PR in WTForms is: https://github.com/wtforms/wtforms/pull/739/files

the change is that "choices" widget expects now 4 values to unpack in `SelectFieldBase.iter_choices` (value, label, selected, render_kw) where previously there were three (value, label, selected). Unfortunately Flask App builder overrides implementation of the `iter_choices` in QuerySelectField https://github.com/dpgaspar/Flask-AppBuilder/blob/38180b0eb54801087157c7bdb3df44a1ed394db3/flask_appbuilder/fields.py#L131   - the `iter_choices` method still returns 3 values which means that when core WTF iterates over the field to display their value here: https://github.com/wtforms/wtforms/pull/739/files#diff-70bc7290d3bd4ce3f5436885709a6b8af1086763e4699598e578ee809027eadbR365 it fails with 

```
>           for val, label, selected, render_kw in field.iter_choices():
E           ValueError: not enough values to unpack (expected 4, got 3)
```

### Environment

Flask-Appbuilder version: Tested with 4.3.6, 4.3.7, 4.3.8 (but previous versions are also affected)
WTForms 3.1.0

pip freeze output:

pip freeze | grep -E "(Flask-App|WTForms)"

```
Flask-AppBuilder==4.3.8
WTForms==3.1.0
```

### Describe the expected results

When instantiating any choice widgets in WTForms 3.1.0 and installing QuerySelectMultipleField 

```
>           for val, label, selected, render_kw in field.iter_choices():
E           ValueError: not enough values to unpack (expected 4, got 3)
```

What should happen - it should not raise the exception.

I am not sure how to easily - minimally - reproduce it but the root cause of the problem is quite clear - It has been detected by the `canary` build in Apache Airflow (when WTForms got automatically upgraded to 3.1.0 which is compatible with FAB).  FAB has WTForms<4, so anyone installing new FAB with latest compatible WTFOrms will have this problem.

[...]

PR #2138, opened 2023-10-15
fix: wtforms compat 3.1.0

dpgaspar:
### Description

fixes: #2137 

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


potiuk (approved):
Yeah. A bit blurry if it's a breaking API or not (WTForms likely did not realising this change might break some users). This seems like a reasonable way to solve it though.

Issue #1641, opened 2021-05-11
AUTH_ROLES_MAPPING not working


cocampbe:
### Environment

Flask-Appbuilder version: 3.3.0

pip freeze output:

```shell
adal==1.2.6
aiohttp==3.7.3
alembic==1.5.4
amqp==2.6.1
apache-airflow==2.0.1
apache-airflow-providers-amazon==1.1.0
apache-airflow-providers-apache-spark==1.0.2
apache-airflow-providers-celery==1.0.1
apache-airflow-providers-cncf-kubernetes==1.0.1
apache-airflow-providers-docker==1.0.1
apache-airflow-providers-elasticsearch==1.0.1
apache-airflow-providers-ftp==1.0.1
apache-airflow-providers-google==1.0.0
apache-airflow-providers-grpc==1.0.1
apache-airflow-providers-hashicorp==1.0.1
apache-airflow-providers-http==1.1.0
apache-airflow-providers-imap==1.0.1
apache-airflow-providers-microsoft-azure==1.1.0
apache-airflow-providers-mysql==1.0.1
apache-airflow-providers-postgres==1.0.1
apache-airflow-providers-redis==1.0.1
apache-airflow-providers-sendgrid==1.0.1
apache-airflow-providers-sftp==1.1.0
apache-airflow-providers-slack==2.0.0
apache-airflow-providers-sqlite==1.0.1
apache-airflow-providers-ssh==1.1.0
apispec==3.3.2
appdirs==1.4.4
argcomplete==1.12.2
async-timeout==3.0.1
attrs==20.3.0
azure-batch==10.0.0
azure-common==1.1.26
azure-core==1.11.0
azure-cosmos==3.2.0
azure-datalake-store==0.0.51
azure-identity==1.5.0
azure-keyvault==4.1.0
azure-keyvault-certificates==4.2.1
azure-keyvault-keys==4.3.1
azure-keyvault-secrets==4.2.0
azure-kusto-data==0.0.45
azure-mgmt-containerinstance==1.5.0
azure-mgmt-core==1.2.2
azure-mgmt-datalake-nspkg==3.0.1
azure-mgmt-datalake-store==0.5.0
azure-mgmt-nspkg==3.0.2
azure-mgmt-resource==15.0.0
azure-nspkg==3.0.2
azure-storage-common==2.1.0
azure-storage-file==2.1.0
Babel==2.9.0
bcrypt==3.2.0
billiard==3.6.3.0
[...]

PR #2154, opened 2023-10-27
feat: partial ldap matching

romsharon98:
### Description
Add an option of partial match between roles

### ADDITIONAL INFORMATION

- [x] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -421,7 +421,7 @@ class BaseSecurityManager(AbstractSecurityManager):
 
     @property
     def auth_partial_matching(self) -> bool:
-        return self.appbuilder.get_app.config.get(["AUTH_PARTIAL_MATCHING"], False)
+        return self.appbuilder.get_app.config.get("AUTH_PARTIAL_MATCHING", False)
 
     @property
     def auth_roles_sync_at_login(self) -> bool:
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -1908,9 +1908,6 @@ class BaseSecurityManager(AbstractSecurityManager):
     def find_role(self, name):
         raise NotImplementedError
 
-    def find_partial_role(self, name):
-        raise NotImplementedError
-
     def add_role(self, name, permissions=None):
         raise NotImplementedError
 
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -638,7 +638,9 @@ class BaseSecurityManager(AbstractSecurityManager):
             log.debug("User info from Azure: %s", me)
             # https://learn.microsoft.com/en-us/azure/active-directory/develop/id-token-claims-reference#payload-claims
             return {
-                "email": me["email"],
+                # To keep backward compatibility with previous versions
+                # of FAB, we use upn if available, otherwise we use email
+                "email": me["upn"] if "upn" in me else me["email"],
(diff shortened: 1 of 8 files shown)


romesharon:
this PR relate to this issue:
https://github.com/dpgaspar/Flask-AppBuilder/issues/1641

flask_appbuilder/security/sqla/manager.py
```diff
@@ -304,9 +304,18 @@ def update_role(self, pk, name: str) -> Optional[Role]:
             return
 
     def find_role(self, name):
-        return (
+        role = (
             self.get_session.query(self.role_model).filter_by(name=name).one_or_none()
         )
+        if role:
+            return role
+        elif self.auth_partial_matching:
```
dpgaspar, line 312: this seems a bit dangerous, since this is a public method that can be used by downstream methods, so fetching a role would change depending on `auth_partial_matching` is unexpected. Also doesn't the like method need wildcards?

dpgaspar (commented):
Thank you for opening this PR, could you please update the description and add more detail about what is the end goal of this PR

romesharon:
The end goal of this PR is to enable partial matching between user ldap groups and AUTH_LDAP_MAPPING.
for example the user belong to those groups:`["OU=7180,DC=f839"]`
and the mapping is:
```python
 AUTH_ROLES_MAPPING = {
            "DC=f839,DC=msl": ["role1"],
            "CN=833/name,OU=Users,OU=787,OU=7180,DC=f839,DC=msl": ["role2"]
          }
```
the user will get `role2` because it partially belong to this mapping.
In this PR the groups need to be in the same order to be partially match, for example if the user belong to this groups: `[OU=Users, OU=7180]` 
it won't get any role

flask_appbuilder/security/sqla/manager.py
romesharon: You are correct about the wildcards I should add %{}% for using like method.
You review gave point to think about and what I did was wrong.
Now I changed my logic and move it the relevant function.

flask_appbuilder/security/manager.py
```diff
@@ -338,17 +338,24 @@ def get_roles_from_keys(self, role_keys: List[str]) -> Set[role_model]:
...
+                _fab_roles.update(fab_role_names)
+        for fab_role_name in _fab_roles:
+            fab_role = self.find_role(fab_role_name)
+            if fab_role:
+                _roles.add(fab_role)
+            else:
+                log.warning(
+                    "Can't find role specified in AUTH_ROLES_MAPPING: %s",
+                    fab_role_name,
+                )
```
dpgaspar, line 358: can you please add a unittest for this method?

tests/test_api.py
```diff
@@ -963,7 +963,7 @@ def test_get_item_not_found(self):
         with model1_data(self.appbuilder.session, 1):
             model_id = MODEL1_DATA_SIZE + 1
             rv = self.auth_client_get(client, token, f"api/v1/model1api/{model_id}")
-            self.assertEqual(rv.status_code, 404)
+            self.assertEqual(rv.status_code, 200)
```
dpgaspar, line 966: strange why this has changed? can you check

goerch: @dpgaspar : he changed it, because intermittently `model_id`  seems to conflict with auto incremented database keys. I'd like to propose the change `model_id = 0` instead.

PR #2168, opened 2023-11-22
feat: Added authentik as new identity provider

Yoyasp:
### Description
Added [Authentik](https://goauthentik.io/) as new identity provider.

This PR includes:
- Authentik as new provider in the security manager.
- Role mapping from Authentik groups
- JWKS validation on Authentik tokens if provided in the config.py.
- New exception if token signature cannot be validated.
- Tests for the authentication for parsing the correct info, checking if the signature is valid and allowing unsigned tokens if specified in the config.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [x] Introduces new feature

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -4,7 +4,7 @@ import re
 from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
 
 from flask import Flask, g, session, url_for
-from flask_appbuilder.exceptions import OAuthProviderUnknown
+from flask_appbuilder.exceptions import OAuthProviderUnknown, InvalidLoginAttempt
 from flask_babel import lazy_gettext as _
 from flask_jwt_extended import current_user as current_user_jwt
 from flask_jwt_extended import JWTManager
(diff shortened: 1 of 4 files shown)


Yoyasp:
Some how flake8 keeps giving me errors in the CI pipeline, however i dont get any errors while running flake8 on my dev host...
Any help is appreciated.

dpgaspar (approved):
LGTM!

flask_appbuilder/security/manager.py
```diff
@@ -701,6 +713,48 @@ def _decode_and_validate_azure_jwt(self, id_token: str) -> Dict[str, str]:
 
         return jwt.decode(id_token, options={"verify_signature": False})
 
+    def _get_authentik_jwks(self, jwks_url) -> dict:
+        import requests
```
dpgaspar, line 717: requests is not an included package, can you change this to use something else like urllib or http.client

dpgaspar (changes requested):
Sorry for the delay, please remove the `requests` dependency

dpgaspar (approved):

dpgaspar:
> Sorry for the delay, please remove the `requests` dependency

forget it we already depend on it, I'll open a PR to include it on extras

flask_appbuilder/security/manager.py
```diff
@@ -701,6 +713,48 @@ def _decode_and_validate_azure_jwt(self, id_token: str) -> Dict[str, str]:
...
+
+        resp = requests.get(jwks_url)
+        if resp.status_code == 200:
+            return resp.json()
+        return False
```
potiuk, line 722: A bit late to the game but I am incorporating latest FAB changes to Airflow. I guess this should be a {} rather than False (or the return type should be `dict | bool`  (but empty dict is likely better as it is Falsey value anyway.

Yoyasp: ...
You are right.

I don't know how I missed that 😧

PR #2235, opened 2024-05-09
fix: Keycloak OAuth2: get groups as role_keys per default

### LINKED ISSUE #1985: Add AUTH_ROLES_MAPPING support using Keycloak OAuth
AUTH_ROLES_MAPPING is not currently not support using the Keycloak provider.

Mimicking what has been done for Azure, it seems we would need to add a single line of code ("role_keys": me.get("roles", []),) in the return dictionary for the provider Keycloak located here:

https://github.com/dpgaspar/Flask-AppBuilder/blob/dcb32c78e6baacb5685843a1e33de4aab9171b80/flask_appbuilder/security/manager.py#L655

berlincount:
### Description

Small change that makes integrating roles with Keycloak dramatically easier, i.e. stops requiring writing custom code.

(fix not feat because this should have been default and makes indirect usability of the library so much easier for users of Superset and the like)

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature

Closes #1985

### DIFF
--- a/flask_appbuilder/security/manager.py
+++ b/flask_appbuilder/security/manager.py
@@ -695,6 +695,7 @@ class BaseSecurityManager(AbstractSecurityManager):
                 "first_name": data.get("given_name", ""),
                 "last_name": data.get("family_name", ""),
                 "email": data.get("email", ""),
+                "role_keys": data.get("groups", []),
             }
         # for Authentik
         if provider == "authentik":


dpgaspar (approved):
Nice! thank you

PR #2277, opened 2024-10-22
fix: hardcoded year removed from copyright

CodeWithEmad:
This will replace the hardcoded year in the copyright part of docs with `datetime.now().year`


dpgaspar (approved):
Thank you!

Issue #2194, opened 2024-02-09
Add support for marshmallow-sqlalchemy 1.0.0


AlphaJack:
### Environment

`pip-freeze`:

```
Flask==2.3.3
Flask-AppBuilder==4.3.11
flask-babel==3.1.0
Flask-Caching==2.0.2
Flask-Compress==1.14
Flask-JWT-Extended==4.5.3
Flask-Limiter==3.5.0
Flask-Login==0.6.3
Flask-Migrate==4.0.5
Flask-OpenID==1.3.0
Flask-Session==0.6.0
Flask-Sockets==0.2.1
Flask-SQLAlchemy==2.5.1
flask-talisman==1.1.0
Flask-WTF==1.1.1
marshmallow==3.20.0
marshmallow-sqlalchemy==1.0.0
```

### Describe the expected results

I am getting a `TypeError` during the initialization of an app that loads Flask-AppBuilder (see below).

I have tried both marshmallow-sqlalchemy 1.0.0, 0.30 and 0.26, but all of them returns the same results.

I am not sure the error involves Flask-AppBuilder directly, but I wanted to check why there is no support for the latest version of marshmallow-sqlalchemy.

### Describe the actual results

When launching the app, I get

```pytb
Traceback (most recent call last):
  File "/usr/bin/superset", line 5, in <module>
    from superset.cli.main import superset
  File "/usr/lib/python3.11/site-packages/superset/__init__.py", line 21, in <module>
    from superset.app import create_app
  File "/usr/lib/python3.11/site-packages/superset/app.py", line 24, in <module>
    from superset.initialization import SupersetAppInitializer
  File "/usr/lib/python3.11/site-packages/superset/initialization/__init__.py", line 28, in <module>
    from flask_appbuilder import expose, IndexView
  File "/usr/lib/python3.11/site-packages/flask_appbuilder/__init__.py", line 5, in <module>
    from .api import ModelRestApi  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/flask_appbuilder/api/__init__.py", line 29, in <module>
    from marshmallow_sqlalchemy.fields import Related, RelatedList
  File "/usr/lib/python3.11/site-packages/marshmallow_sqlalchemy/__init__.py", line 1, in <module>
    from .schema import (
[...]

PR #2298, opened 2025-01-08
fix: relax marshmallow-sqlalchemy version constraint to fix compatibility with marshmallow>=3.24

sloria:
### Description

Relax the upper marshmallow-sqlalchemy version constraint. This is necessary for compatibility with newer versions of marshmallow (>=3.24.0).

This would resolve this issue in Airflow: https://github.com/apache/airflow/pull/45442#issuecomment-2578324005

This also upgrades a few dependencies needed for a working development environment (used `pip-compile --upgrade-package ...`).

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/dpgaspar/Flask-AppBuilder/issues/2194
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


sloria:
given that marshmallow 4.0 will be released soon, i also think the version constraint for marshmallow should also be relaxed to `marshmallow>=3.18.0, <4`. upon quick review, i didn't see any usages within the code that would be broken by the changes in 4.0.

but this can happen as a follow-up

potiuk:
Nice one - would be great to get it in :)

setup.py
```diff
@@ -58,7 +58,7 @@ def desc():
         "Flask-JWT-Extended>=4.0.0, <5.0.0",
         "jsonschema>=3, <5",
         "marshmallow>=3.18.0, <4",
-        "marshmallow-sqlalchemy>=0.22.0, <0.29.0",
+        "marshmallow-sqlalchemy>=0.22.0, <2.0.0",
```
sloria, line 61: IMO we should remove the upper bound altogether

```suggestion
        "marshmallow-sqlalchemy>=0.22.0",
```

to avoid issues like the one that Airflow ran into. since FAB is a library, its shouldn't impose such strict upper constraints on its dependencies and allow users to pin their own dependencies

sloria:
i've relaxed the versions for both marshmallow and marshmallow-sqlachemy even further because there are major releases planned for both but neither should affect FAB

dpgaspar (approved):
Nice! thank you

potiuk:
Cool!

PR #2321, opened 2025-03-14
fix: Ensure sub is a string when generating JWT tokens

withnale:
### Description

There has been a recent change in PyJWT and this library has become less forgiving.

Specifically, it will now fail to decode JWT tokens if the sub field is not a string, generating an error of the form `Subject must be a string`.

Whilst it would be ideal to fix this at source, there are likely to be lots of applications that rely on this code where `user.id`
already exists and is not a string. If feels like the simplest solution is to ensure user.id is cast to a string whenever used as part of JWT token generation.

### ADDITIONAL INFORMATION

- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [ ] Introduces new feature


Benjizle (approved):

dpgaspar (approved):
Thank you for the fix

PR #2331, opened 2025-04-04
fix(translations): Update messages.po for French translation

xavier-GitHub76:
Complete review of translations.
This file is in line with the POT of 04/04/2025


xavier-GitHub76:
My previous message might be misleading.
This PO file is built from POT (2024-10-22 12:26+0330)

PR #2451, opened 2026-04-08
feat(security): Support POST OAuth callback endpoint

LipuFei:
### Description

This PR adds an HTTP `POST` method for the `/oauth-authorized/<provider>` callback endpoint.

Microsoft Azure supports `response_mode=fom_post` which uses a HTTP POST callback. This is more secure than GET with the state code in the URL itself.

See https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-authorization-code

**NOTE:** When the POST endpoint is used with Flask-WTF for CSRF protection, this OAuth2 callback endpoint must be exempted from CSRF or the Flask session cookie will not be available, and the callback endpoint won't be able to find the corresponding session to complete the authentication flow.

### ADDITIONAL INFORMATION
- [ ] Has associated issue:
- [ ] Is CRUD MVC related.
- [x] Introduces new feature

### DIFF
--- a/flask_appbuilder/baseviews.py
+++ b/flask_appbuilder/baseviews.py
@@ -119,6 +119,11 @@ class BaseView(AbstractViewApi):
     route_base = None
     """ Override this if you want to define your own relative url """
 
+    csrf_exempt = False
+    """
+    If using flask-wtf CSRFProtect exempt the View from check
+    """
+
     template_folder = "templates"
     """ The template folder relative location """
     static_folder = "static"
@@ -299,6 +304,13 @@ class BaseView(AbstractViewApi):
                 static_folder=static_folder,
             )
         self._register_urls()
+
+        # Exempt from CSRF if the view opts in
+        if getattr(self, "csrf_exempt", False):
+            csrf = current_app.extensions.get("csrf")
+            if csrf:
+                csrf.exempt(self.blueprint)
+
         return self.blueprint
 
     def _register_urls(self):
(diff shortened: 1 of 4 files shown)


dpgaspar (commented):
Let's exempt this endpoint then, we can do the following:
``` python
  # In BaseView.create_blueprint (baseviews.py:262)
  def create_blueprint(self, appbuilder, endpoint=None, static_folder=None):
      ...
      self._register_urls()
      # Exempt from CSRF if the view opts in
      if getattr(self, "csrf_exempt", False):
          csrf = current_app.extensions.get("csrf")
          if csrf:
              csrf.exempt(self.blueprint)
      return self.blueprint

  Then the PR just needs:

  class AuthOAuthView(AuthView):
      csrf_exempt = True
```

Could you please add tests for POST

LipuFei:
@dpgaspar Thanks for the review. I will try to add a test for POST.

For `csrf_exempt = True`, I'm a little hesitant to exempt the complete `AuthOAuthView`. The POST callback is the only one that needs to be exempted to use `response_mode=fom_post`. I have tested this on a modified Airflow deployment.

I personally would prefer to keep the CSRF exemption scope small. Is there a way to just exempt `oauth_authorized()` in this case?

LipuFei:
Did what you suggested. Added a unit test and added `csrf_exempt` to `BaseView`.

